PR #2, opened 2014-07-04
7.0 fix flakes8

yvaucher:


base_report_to_printer/report_service.py
```diff
@@ -41,22 +42,26 @@ def exp_report_get(self, db, uid, report_id):
...
+                        report_obj.print_direct(
+                            cr, uid, report.id, base64.encodestring(self._reports[report_id]['result']),
                             self._reports[report_id]['format'], printer)
                         # XXX "Warning" removed as it breaks the workflow
                         # it would be interesting to have a dialog box to confirm if we really want to print
                         # in this case it must be with a by pass parameter to allow massive impression
-                        #raise osv.except_osv(_('Printing...'), _('Document sent to printer %s') % (printer,))
+                        # raise osv.except_osv(
+                        #     _('Printing...'),
+                        #     _('Document sent to printer %s') % (printer,))
```
bwrsandman, line 64: You could probably get rid of the commented-out code, no?

bwrsandman:
:+1:

Looks a lot better!

base_report_to_printer/report_service.py
yvaucher, line 64: I'm not convinced with this is yet. There are no proper message send to the user.

I let the XXX to know there is something to fix here code is more or less part of the comment as the comment explain why we don't raise anything.

bwrsandman, line 64: I usually got with `FIXME` instead of `XXX` as a lot of editors will highlight that string along with `TODO`

yvaucher, line 64: XXX will also be recognized and highlighted but ok FIXME is more descriptive

PR #9, opened 2014-10-24
[IMP] remove useless locking from update()

clonedagain:
Locking shouldn't be needed here because the data we read is immutable.
Supersedes https://code.launchpad.net/~numerigraphe-team/report-print-send/7.0-no-lock-in-update-1308635-ls/+merge/216353
"If I'm not mistaken, locking in printer.printer.update() is not needed, and the way it's written does not protect us from concurrent changes anyway :
- if what we read is immutable than the value is copied and locking is unnecessary
- if it's mutable then we're getting a "reference" to the shared object, and should keep the lock until we're completely done with it

From my quick code survey, the variables seem to contain a boolean and a float so I guess no locking is needed.

update() seems to be called pretty often (all the CRUD method call it) so It may help a bit to remove the contention."

### DIFF
--- a/base_report_to_printer/printing.py
+++ b/base_report_to_printer/printing.py
@@ -155,23 +155,19 @@ class printing_printer(orm.Model):
         thread.start()
 
     def update(self, cr, uid, context=None):
-        if not context or context.get('skip_update'):
-            return
-        self.lock.acquire()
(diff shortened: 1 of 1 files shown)


guewen:
:+1:

pedrobaeza:
:+1:

PR #12, opened 2014-11-14
Migration of base_report_to_printer

guewen:
# Migration of base_report_to_printer
- Migration to the new API
- Compatible with the new reports (`report` module, but only the `pdf` ones, not implemented for `html` reports, I don't think that's useful)
- Compatible with deprecated reports (rml, webkit, ...)
- Removed the `Lock()` local to the process and replaced it by a database lock so the processes share the same lock and last update timestamp
- Refreshment of the views


base_report_to_printer/users.py
```diff
@@ -22,27 +21,23 @@
...
+                in _available_action_types(self)
+                if code != 'user_default']
 
-    _columns = {
-        'printing_action': fields.selection(_user_available_action_types, 'Printing Action'),
-        'printing_printer_id': fields.many2one('printing.printer', 'Default Printer'),
-        }
-
-# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
+    printing_action = fields.Selection(_user_available_action_types)
```
yvaucher, line 41: Field string 'Printing Action' was removed intentionally ?

guewen, line 41: Yes.
As advised in https://www.odoo.com/documentation/8.0/reference/orm.html in **Porting from the old API**:

> if a field’s string is the titlecased version of the field name:
> `name = fields.Char(string="Name")`
> it is useless and should be removed

yvaucher, line 41: ok

yvaucher:
Great cleaning and thanks for the job done here :+1:

guewen:
I will make a few changes (extract a method at least) to reduce the code in printer_tray.

guewen:
Done.

yvaucher:
There is an issue on install: Programming error: relation "printing_printer_polling" does not exist

guewen:
WTF

@yvaucher I fixed it

yvaucher:
:+1:

guewen:
but my last commit is bad

guewen:
Pushed the fix.

pedrobaeza:
:+1: 

Thanks for the conversion to new API.

PR #13, opened 2014-11-17
Migration of printer_tray to 8.0

guewen:
Migrated the module `printer_tray`.


yvaucher:
:+1:

printer_tray/README.rst
```diff
@@ -0,0 +1,15 @@
...
+*This module extends `Report to printer` module.*
+
+It detects trays on printer installation plus permits to select the
+paper source on which you want to print directly.
+
+You will find this option on default user config, on default report
+config and on specific config per user per report.
+
+This allows you to dedicate a specific paper source for exemple for
+preprinted paper such as payment slip.
```
pedrobaeza, line 14: You should consider to add OCA description template

printer_tray/README.rst
```diff
@@ -0,0 +1,15 @@
...
+
+*This module extends `Report to printer` module.*
+
+It detects trays on printer installation plus permits to select the
+paper source on which you want to print directly.
+
+You will find this option on default user config, on default report
+config and on specific config per user per report.
+
+This allows you to dedicate a specific paper source for exemple for
```
pedrobaeza: s/exemple/example

printer_tray/users.py
```diff
@@ -18,15 +18,15 @@
...
+class ResUsers(models.Model):
 
     _inherit = "res.users"
 
-    _columns = {
-        'printer_tray_id': fields.many2one(
-            'printing.tray', 'Default Printer Paper Source',
-            domain="[('printer_id', '=', printing_printer_id)]"),
-        }
+    printer_tray_id = fields.Many2one(
```
pedrobaeza, line 28: You should put an onchange on printer to select the first available tray from new printer

printer_tray/report_xml_action.py
```diff
@@ -19,20 +19,21 @@
...
-class ReportXMLAction(orm.Model):
+class ReportXMLAction(models.Model):
     _inherit = 'printing.report.xml.action'
 
-    _columns = {
-        'printer_tray_id': fields.many2one(
-            'printing.tray', 'Paper Source',
-            domain="[('printer_id', '=', printer_id)]"),
-        }
+    printer_tray_id = fields.Many2one(
```
pedrobaeza, line 28: You should put an onchange on printer to select the first available tray from new printer

printer_tray/ir_report.py
```diff
@@ -0,0 +1,32 @@
...
+#
+##############################################################################
+from openerp import models, fields
+
+
+class IrActionsReportXml(models.Model):
+
+    _inherit = 'ir.actions.report.xml'
+
+    printer_tray_id = fields.Many2one(
```
pedrobaeza, line 28: You should put an onchange on printer to select the first available tray from new printer

printer_tray/users.py
guewen, line 28: No,  the tray field can be empty and means that cups will print on the default tray of the printer. It is preferable to leave the field empty rather than copy this value from the cups server, because if the default tray is changed on the cups server,  the tray stored in the records would no longer be correct.

printer_tray/report_xml_action.py
guewen, line 28: No,  the tray field can be empty and means that cups will print on the default tray of the printer. It is preferable to leave the field empty rather than copy this value from the cups server, because if the default tray is changed on the cups server,  the tray stored in the records would no longer be correct.

printer_tray/ir_report.py
guewen, line 28: No, the tray field can be empty and means that cups will print on the default tray of the printer. It is preferable to leave the field empty rather than copy this value from the cups server, because if the default tray is changed on the cups server, the tray stored in the records would no longer be correct.

printer_tray/README.rst
guewen, line 14: Done

printer_tray/ir_report.py
pedrobaeza, line 28: Well, then at least empty it, because if you have one tray selected from another printer, and you change the printer, this tray remains selected.

guewen, line 28: That's right. I'm just searching a bit, I have a tray that is called "Auto" so it could be selected but I'm not sure if I have a proper way to identify it (perhaps it is always called "Auto"?)

pedrobaeza, line 28: Not always, because it depends on the corresponding PPD... The best bet is just to leave it empty as you have suggested.

guewen, line 28: Umm anyway, we still need to put the field to blank at least on the report and report per user because it fallbacks on the user's config if empty. Still I need to empty the field on printer's changes.

pedrobaeza, line 28: Yeah, it seems so.

guewen, line 28: Updated: the tray is now reset when the printer is changed

pedrobaeza:
Thanks :+1: 

When Travis goes green, I'll merge.

guewen:
Thanks @pedrobaeza

PR #15, opened 2014-11-20
Use a cron instead of threads to update printers status

### LINKED ISSUE #14: 8.0 base_report_to_printer blocks HTTP processes in multiprocess
When the 8.0 version of base_report_to_printer is used, the server can't start in multiprocess mode. No issues when `--workers=0`.

It seems related to the threading.

guewen:
The implementation with threads was blocking the loading of the server in
multiprocess.  Using a cron will lower the frequency of the updates but at
least it is simple and reliable.

Fixes #14

### DIFF
--- a/base_report_to_printer/printing.py
+++ b/base_report_to_printer/printing.py
@@ -24,194 +24,13 @@
 import logging
 import os
 
-from contextlib import contextmanager
-from datetime import datetime
 from tempfile import mkstemp
-from threading import Thread
 
 import cups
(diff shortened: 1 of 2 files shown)


pedrobaeza:
What a cleaning! Indeed, this doesn't provide real time, but it's safer and don't include possible lockings.

:+1:

alexis-via:
Big code simplification ! :+1:

yvaucher:
:+1:

PR #18, opened 2014-12-19
base_report_to_printer: missing security

guewen:
Users are not able to print or to generate a pdf

### DIFF
--- a/base_report_to_printer/security/security.xml
+++ b/base_report_to_printer/security/security.xml
@@ -42,5 +42,21 @@
       <field eval="0" name="perm_write"/>
       <field eval="0" name="perm_create"/>
     </record>
+    <record id="ir_model_access_printing_action_all0" model="ir.model.access">
+      <field name="name">printing_action all</field>
+      <field name="model_id" ref="model_printing_action"/>
+      <field eval="1" name="perm_read"/>
+      <field eval="0" name="perm_unlink"/>
+      <field eval="0" name="perm_write"/>
+      <field eval="0" name="perm_create"/>
+    </record>
+    <record id="ir_model_access_printing_report_xml_action_all0" model="ir.model.access">
+      <field name="name">printing_report_xml_action all</field>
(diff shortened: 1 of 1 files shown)


base_report_to_printer/security/security.xml
```diff
@@ -1,46 +1,62 @@
...
-                </record>
+  <data noupdate="1">
+    <record id="res_groups_printingprintoperator0" model="res.groups">
+      <field name="name">Printing / Print Operator</field>
+    </record>
+    <record id="ir_model_access_printingprintergroup1" model="ir.model.access">
+      <field name="name">printing_printer group</field>
+      <field name="model_id" ref="model_printing_printer"/>
+      <field name="group_id" ref="res_groups_printingprintoperator0"/>
+      <field eval="1" name="perm_read"/>
```
pedrobaeza, line 11: Why don't you put?: `<field name="perm_read">1</field>` or better, make a CSV file that it's more comfortable.

guewen, line 11: Because it already was a XML, I didn't want to change that (and the fact that it has a noupdate section, changing that to a CSV that is always updated may surprise users).

For boolean fields, I always use `eval` (or in this case I kept `eval`). This is what the docs says: 

> This will evaluate to the string 'False' and not the boolean False. This is especially tricky because Python's conversion rules consider any non-empty string to be True, so the above code will end up storing the opposite of what is desired.
> If you want to evaluate the value to a float, a boolean or another type, except string, you need to use the eval attribute:

Ref: https://doc.odoo.com/trunk/server/03_module_dev_03/

pedrobaeza, line 11: Thanks for the info.

pedrobaeza:
:+1: 

Good to see OCA runbot making their first steps!

eLBati:
:+1:

alexis-via:
:+1: Code review and test
In fact, I was impacted by this bug when deploying the module at my customer site... and then found the fix here !

Issue #19, opened 2014-12-28
8.0 : base_report_to_printer fails when printing with option "raw"


alexis-via:
With an up-to-date code of the 8.0 branch, when I execute the following line of code:

```
self.pool['printing.printer'].print_document(cr, uid, printer_id, 'colissimo.zpl', zpl_file, 'raw', context=context)
```

I get a crash, which is linked to the "raw" option:

```
2014-12-28 00:33:10,091 15416 ERROR o8_barroux8 openerp.http: Exception during JSON request handling.
Traceback (most recent call last):
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 518, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 539, in dispatch
    result = self._call_function(**self.params)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 295, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/service/model.py", line 113, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 292, in checked_call
    return self.endpoint(*a, **kw)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 755, in __call__
    return self.method(*args, **kw)
  File "/home/alexis/new_boite/dev/odoo/openerp/http.py", line 388, in response_wrap
    response = f(*args, **kw)
  File "/home/alexis/new_boite/dev/odoo/addons/web/controllers/main.py", line 953, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/alexis/new_boite/dev/odoo/addons/web/controllers/main.py", line 941, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
  File "/home/alexis/new_boite/dev/odoo/openerp/api.py", line 237, in wrapper
    return old_api(self, *args, **kwargs)
  File "/home/alexis/new_boite/dev/barroux/barroux_stock/stock.py", line 146, in print_picking_wave_label_zpl
    zpl_file, 'raw', context=context)
  File "/home/alexis/new_boite/dev/odoo/openerp/api.py", line 237, in wrapper
    return old_api(self, *args, **kwargs)
[...]

PR #20, opened 2014-12-28
[FIX] Bug #19 avoid crash when using 'raw' option

alexis-via:


pedrobaeza:
Curious patch

:+1:

clonedagain:
:+1: it most probably derives from the keyword 'True' which enables features in the Adobe PPD specifications.

guewen:
:+1:

PR #22, opened 2015-01-08
base_report_to_printer: add support for remote CUPS server (not just localhost)

alexis-via:
More logging and better error handling

### DIFF
--- a/base_report_to_printer/printing.py
+++ b/base_report_to_printer/printing.py
@@ -144,12 +144,14 @@ class PrintingPrinter(models.Model):
 
         options = self.print_options(report, format)
 
-        _logger.debug('Sending job to CUPS printer %s' % self.system_name)
+        _logger.debug(
+            'Sending job to CUPS printer %s on %s'
+            % (self.system_name, CUPS_HOST))
         connection.printFile(self.system_name,
                              file_name,
                              file_name,
                              options=options)
-        _logger.info("Printing job: '%s'" % file_name)
+        _logger.info("Printing job: '%s' on %s" % (file_name, CUPS_HOST))
         return True
 
     @api.multi
--- a/base_report_to_printer/printing.py
+++ b/base_report_to_printer/printing.py
@@ -23,14 +23,15 @@
 ##############################################################################
 import logging
 import os
-
 from tempfile import mkstemp
-
 import cups
-
-from openerp import models, fields, api
+from openerp import models, fields, api, _
+from openerp.exceptions import Warning
+from openerp.tools.config import config
 
 _logger = logging.getLogger(__name__)
+CUPS_HOST = config.get('cups_host', 'localhost')
+CUPS_PORT = int(config.get('cups_port', 631))  # config.get returns a string
 
 
 class PrintingPrinter(models.Model):
@@ -63,7 +64,7 @@ class PrintingPrinter(models.Model):
     def update_printers_status(self):
         printer_recs = self.search([])
         try:
-            connection = cups.Connection()
+            connection = cups.Connection(CUPS_HOST, CUPS_PORT)
             printers = connection.getPrinters()
         except:
             printer_recs.write({'status': 'server-error'})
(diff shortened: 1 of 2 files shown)


alexis-via:
One of the important things in this MP is : where do we store the host + port of the CUPS server ? I chose to store it in the server config file, because I find it simple and easy. Of course, we could store in an ir_config param, on a field of res.company, etc...

rvalyi:
IMHO, config in database is mostly relevant only when you have a multi-db installation. But here as the OpenERP server will need to connect to the cups server directly, I think there will never be a situation where one db would want to connect to a given CUPS and another db would connect to another CUPS.

Hence I agree with your choice to have that config simply in the damn config file.

base_report_to_printer/printing.py
```diff
@@ -127,10 +128,23 @@ def print_document(self, report, content, format):
...
+        try:
+            _logger.debug(
+                'Starting to connect to CUPS on %s:%s'
+                % (CUPS_HOST, CUPS_PORT))
+            connection = cups.Connection(CUPS_HOST, CUPS_PORT)
+            _logger.debug('Connection to CUPS successfull')
+        except:
+            raise Warning(
+                _("Failed to connect to the CUPS server on %s:%s. "
+                    "Check that the CUPS server is running and that "
```
pedrobaeza, line 141: Incorrect indentation

pedrobaeza:
One minor issue, the rest :+1:

clonedagain:
:+1: except for what @pedrobaeza already reported.
Very useful feature, since the easiest option to get up-to-date print drivers is to upgrade the whole distribution, and we don't want to do that on the odoo server.

base_report_to_printer/printing.py
```diff
@@ -127,10 +128,23 @@ def print_document(self, report, content, format):
...
+                    "you can reach it from the Odoo server.")
+                % (CUPS_HOST, CUPS_PORT))
 
         options = self.print_options(report, format)
 
+        _logger.debug('Sending job to CUPS printer %s' % self.system_name)
```
clonedagain: Nitpicking: maybe it's worth mentioning the values of CUPS_HOST and CUPS_PORT here, so that we don't go searching for jobs on the wrong host?

base_report_to_printer/printing.py
alexis-via, line 141: @pedrobaeza I don't understand your point about the incorrect indentation. It is flake8 compliant, and I just put the "cups.Connection" inside a try/except with some logs.

clonedagain, line 141: @alexis-via it's just about the string, when we break them into several lines  we usually align all lines to the same indent as the first `"`.

pedrobaeza, line 141: Indeed I'm surprised that flake8 allows that strange indentation. I prefer to align all, but it's not a blocking issue.

yvaucher:
:+1: Thanks

clonedagain:
Thanks for this merge.
I'm trying to make use of this feature but it seems that python-cups can get the list of printers on some of my remote cups, but not on others.
I tried to copy the cupsd.conf of one server on the other, or to enable printer sharing with cupsctl but I had no success.
Does one of you have an idea by any chance?

A session on a server which doesn't works:

```
Python 2.7.8 (default, Oct 20 2014, 15:05:19)
[GCC 4.9.1] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import cups
>>> cnx=cups.Connection('alpha-cupserp')
>>> cnx.getPrinters()
{}
```

A session on a server which works, initiated from the same client:

```
>>> cnx=cups.Connection('10.1.4.6')
>>> cnx.getPrinters()
{u'Zebra-Technologies-QLn320': {'printer-is-shared': True, 'printer-info': u'Zebra Technologies QLn320', 'printer-state-message': u'', 'printer-type': 134254660, 'printer-make-and-model': u'Zebra ZPL Label Printer', 'printer-state-reasons': [u'none'], 'printer-uri-supported': u'ipp://10.1.4.6:631/printers/Zebra-Technologies-QLn320', 'printer-state': 3, 'printer-location': u'soho', 'device-uri': u'usb://Zebra%20Technologies/QLn320?serial=QLn320'}, u'Photosmart_C4500': {'printer-is-shared': True, 'printer-info': u'Photosmart_C4500', 'printer-state-message': u'', 'printer-type': 134385692, 'printer-make-and-model': u'HP Photosmart c4500 Series, hpcups 3.14.6', 'printer-state-reasons': [u'none'], 'printer-uri-supported': u'ipp://10.1.4.6:631/printers/Photosmart_C4500', 'printer-state': 3, 'printer-location': u'', 'device-uri': u'hp:/net/Photosmart_C4500_series?ip=192.168.1.36'}}
```

PR #23, opened 2015-01-21
Do no longer returns a PDF when a report is printed

### LINKED ISSUE #16: base_report_to_printer opens a copy of sent report
An improvement would be to avoid opening the report.

Previously we did an attempt to raise a warning message to give a feed back to the user that the report has been printed. But this was breaking the workflow when printing multiple reports. So it needs some improvements.

guewen:
Instead, a notification is displayed to the user. When report.get_pdf() is
called on a report that must be printer, it will print the report _and_
returns the pdf, thus code that calls directly report.get_pdf() will print
the pdf on the printer as expected.

Fixes #16

### DIFF
--- a/base_report_to_printer/README.rst
+++ b/base_report_to_printer/README.rst
@@ -5,7 +5,7 @@ This module allows users to send reports to a printer attached to the server.
 
 It adds an optional behaviour on reports to send it directly to a printer.
 
-* `Send to Client` is the default behavious providing you a downloadable PDF
+* `Send to Client` is the default behaviour providing you a downloadable PDF
 * `Send to Printer` prints the report on selected printer
 
 Report behaviour is defined by settings.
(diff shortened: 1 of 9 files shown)


guewen:
Notification when a document is sent to the printer

pedrobaeza:
Excellent addition!

Fast code review: :+1:

yvaucher:
That a need to have.
Thanks!

:+1:

guewen:
One limitation: it will not show the notification on the old rml reports, only on the qweb-pdf reports. (but there is no change for them, they are still printed on the printer and not downloaded, there is just no notification, this is because they use `exp_report_get`)

yvaucher:
What about a notice about it in description?

guewen:
Done

idisgc84:
Hello, I'm having the same problem you're describing here. For QWeb reports the module sends them directly to the printer without saving a local copy but for Jasper Reports it saves a copy and that isn't the behavior we want.

PR #25, opened 2015-01-30
Compatibility fix for report with custom parser

nbessi:
By calling `super.get_pdf` in print_document we can encounter trouble with MRO resolution
that prevent custom report parser (e.g. override of `get_pdf`) to be called.

The fix consist of not calling `super` and prevent multiple call to 'printer.print_document'

### DIFF
--- a/base_report_to_printer/report.py
+++ b/base_report_to_printer/report.py
@@ -50,9 +50,7 @@ class Report(models.Model):
         If you want to prevent `get_pdf` to send report you can set
         the `must_skip_send_to_printer` key to True in the context
         """
-        if context is None:
-            context = self.pool['res.users'].context_get(cr, uid)
-        if context.get('must_skip_send_to_printer'):
+        if context is not None and context.get('must_skip_send_to_printer'):
             return False
         if behaviour['action'] == 'server' and printer and document:
             return True
--- a/base_report_to_printer/report.py
+++ b/base_report_to_printer/report.py
@@ -25,12 +25,26 @@ from openerp import models, exceptions, _
 class Report(models.Model):
     _inherit = 'report'
 
+    def _can_send_report(self, cr, uid, ids, behaviour, printer, document,
+                         context=None):
+        """Predicate that decide if report can be sent to printer
+
+        If you want to prevent `get_pdf` to send report you can set
+        the `must_skip_sent_to_printer` key to True in the context
+        """
+        if context is None:
+            context = self.pool['res.users'].context_get(cr, uid)
+        if context.get('must_skip_sent_to_printer'):
+            return False
+        if behaviour['action'] == 'server' and printer and document:
+            return True
+        return False
+
     def print_document(self, cr, uid, ids, report_name, html=None,
                        data=None, context=None):
         """ Print a document, do not return the document file """
-        document = super(Report, self).get_pdf(cr, uid, ids, report_name,
-                                               html=html, data=data,
-                                               context=context)
+        document = self.get_pdf(cr, uid, ids, report_name,
+                                html=html, data=data, context=context)
         report = self._get_report_from_name(cr, uid, report_name)
         behaviour = report.behaviour()[report.id]
         printer = behaviour['printer']
@@ -47,12 +61,19 @@ class Report(models.Model):
         If the action configured on the report is server, it prints the
         generated document as well.
         """
+        if context is None:
+            context = self.pool['res.users'].context_get(cr, uid)
         document = super(Report, self).get_pdf(cr, uid, ids, report_name,
                                                html=html, data=data,
                                                context=context)
         report = self._get_report_from_name(cr, uid, report_name)
         behaviour = report.behaviour()[report.id]
         printer = behaviour['printer']
-        if behaviour['action'] == 'server' and printer and document:
-            printer.print_document(report, document, report.report_type)
+        can_send_report = self._can_send_report(cr, uid, ids,
+                                                behaviour, printer, document,
+                                                context=context)
+        if can_send_report:
+            sent = printer.print_document(report, document, report.report_type)
+            context['must_skip_sent_to_printer'] = True
+            return sent
         return document
--- a/base_report_to_printer/report.py
+++ b/base_report_to_printer/report.py
@@ -43,8 +43,12 @@ class Report(models.Model):
     def print_document(self, cr, uid, ids, report_name, html=None,
                        data=None, context=None):
         """ Print a document, do not return the document file """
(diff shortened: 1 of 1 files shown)


base_report_to_printer/report.py
```diff
@@ -25,12 +25,26 @@
...
+        return False
+
     def print_document(self, cr, uid, ids, report_name, html=None,
                        data=None, context=None):
         """ Print a document, do not return the document file """
-        document = super(Report, self).get_pdf(cr, uid, ids, report_name,
-                                               html=html, data=data,
-                                               context=context)
+        document = self.get_pdf(cr, uid, ids, report_name,
+                                html=html, data=data, context=context)
```
guewen: `get_pdf` should be called with the must_skip_sent_to_printer key to avoid to be printed 2 times

base_report_to_printer/report.py
```diff
@@ -47,12 +61,19 @@ def get_pdf(self, cr, uid, ids, report_name, html=None,
...
         printer = behaviour['printer']
-        if behaviour['action'] == 'server' and printer and document:
-            printer.print_document(report, document, report.report_type)
+        can_send_report = self._can_send_report(cr, uid, ids,
+                                                behaviour, printer, document,
+                                                context=context)
+        if can_send_report:
+            sent = printer.print_document(report, document, report.report_type)
+            context['must_skip_sent_to_printer'] = True
+            return sent
```
guewen: `get_pdf` must returns the pdf document, so the return here is to remove.
Setting the key in the context here has no effect?

base_report_to_printer/report.py
```diff
@@ -47,12 +65,18 @@ def get_pdf(self, cr, uid, ids, report_name, html=None,
...
         report = self._get_report_from_name(cr, uid, report_name)
         behaviour = report.behaviour()[report.id]
         printer = behaviour['printer']
-        if behaviour['action'] == 'server' and printer and document:
+        can_send_report = self._can_send_report(cr, uid, ids,
+                                                behaviour, printer, document,
+                                                context=context)
+        if can_send_report:
             printer.print_document(report, document, report.report_type)
+            context['must_skip_sent_to_printer'] = True
```
guewen: Can you remove this line? (and so the lines 68-69 as well)

guewen: Fixed in https://github.com/nbessi/report-print-send/pull/1

base_report_to_printer/report.py
```diff
@@ -40,6 +43,21 @@ def print_document(self, cr, uid, ids, report_name, html=None,
...
+        the `must_skip_send_to_printer` key to True in the context
+        """
+        if context is None:
+            context = self.pool['res.users'].context_get(cr, uid)
+        if context.get('must_skip_send_to_printer'):
+            return False
+        if behaviour['action'] == 'server' and printer and document:
+            return True
+        return False
+
```
clonedagain, line 58: You're not using anything from the user's context, no need to get it.
The logic is confusing. How about changing the method body to a more direct form:

```
if context and context.get('must_skip_send_to_printer'):
    return False
else:
    return behaviour['action'] == 'server' and printer and document
```

base_report_to_printer/report.py
```diff
@@ -53,6 +71,9 @@ def get_pdf(self, cr, uid, ids, report_name, html=None,
         report = self._get_report_from_name(cr, uid, report_name)
         behaviour = report.behaviour()[report.id]
         printer = behaviour['printer']
-        if behaviour['action'] == 'server' and printer and document:
+        can_print_report = self._can_print_report(cr, uid, ids,
+                                                  behaviour, printer, document,
+                                                  context=context)
+        if can_print_report:
```
clonedagain, line 75: How about skipping the var:

```
if self._can_print_report(cr, uid, ids,
                          ...
```

base_report_to_printer/report.py
```diff
@@ -28,9 +28,12 @@ class Report(models.Model):
     def print_document(self, cr, uid, ids, report_name, html=None,
                        data=None, context=None):
         """ Print a document, do not return the document file """
-        document = super(Report, self).get_pdf(cr, uid, ids, report_name,
-                                               html=html, data=data,
-                                               context=context)
+        if context is None:
+            context = self.pool['res.users'].context_get(cr, uid)
```
clonedagain, line 32: Do you need user context here? would the rest of the code miss something if you just used `{}`?

clonedagain:
Would you care to squash the commits, and maybe consider my nitpicking remarks above?

Overall :+1:, reading the code the change looks sane to me. I didn't test, sorry, I wouldn't know how to do easily.
Do you have a concrete example that you could turn into a unit test by any chance?

base_report_to_printer/report.py
nbessi, line 32: @clonedagain please refer to mailing list https://lists.launchpad.net/openerp-community/msg06816.html.

base_report_to_printer/report.py
nbessi, line 58: I would prefer

```
if context is not None and context.get('must_skip_send_to_printer'):
```

base_report_to_printer/report.py
nbessi, line 75: If found it less readable matter of taste

nbessi, line 75: @guewen I wonder if we should not restore the context modification with `must_skip_send_to_printer` in case someone does something mutliple/bad overriding of `get_pdf`. It will add some defensiveness to the code.

guewen, line 75: I don't get it.

nbessi, line 75: I was talking about https://github.com/nbessi/report-print-send/commit/05f023e18b437a8614b84ce1d43a78f79e3f987d

guewen, line 75: The alteration of the context was pointless: the context was never used afterwards.
If your idea was to change it so it goes up to the caller, I think it's anyway a bad idea and that's the caller's work to ensure it calls the method correctly. By the way such alteration up to the caller won't be possible with the new API where the context is a frozendict.

nbessi, line 75: @guewen my point was more about mro trouble caused by dependency issues on parallel addons that may cause multiple call to printer. But ok lets see if the bug happens and we will advise then

base_report_to_printer/report.py
clonedagain, line 58: More foolproof, agreed.

guewen:
:+1:

guewen:
Rebased and merged in 8.0

PR #28, opened 2015-05-21
[8.0] [FIX]: change signature of methods print_document & get_pdf for 8.0 api

anthony-muschang:
It is currently not possible to do the following calls:

``` python
report_model = self.env['report']
invoice_model = self.env['account.invoice']

invoices = invoice_model.search([])
report_model.get_pdf(invoices, 'account.report_invoice')
report_model.print_document(invoices, 'account.report_invoice')
```

The signature of the methods print_document & get_pdf have been modified in order to be able to call them from 8.0 api.

I use the same signature as seen in the standard 'report' module:
https://github.com/odoo/odoo/blob/453b373a685d652bc8e542284a0da899ce5ffccd/addons/report/models/report.py#L277

### DIFF
--- a/base_report_to_printer/report.py
+++ b/base_report_to_printer/report.py
@@ -25,6 +25,7 @@ from openerp import models, exceptions, _, api
 class Report(models.Model):
     _inherit = 'report'
 
+    @api.v7
     def print_document(self, cr, uid, ids, report_name, html=None,
                        data=None, context=None):
         """ Print a document, do not return the document file """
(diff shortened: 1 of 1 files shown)


guewen:
Makes sense.
Is the change necessary for `get_pdf` since it is already done in the core?

anthony-muschang:
Yes. Because the get_pdf of base_report_to_printer will be call first.

guewen:
Indeed, but I thought the decorators were inherited and thus it was not necessary to repeat them. It is maybe not true for `@api.v7` and `@api.v8`.
:+1:

pedrobaeza:
Please make it the other way and let the main method to be @api.v8, so the migration to v9 will be smoother.

yvaucher:
:+1:

lmignon:
:+1:

pedrobaeza:
Don't merge until my comment is reviewed

guewen:
@pedrobaeza considering that:
- the base model is not migrated
- the `report` model is a mess to use with the new API because in this model `ids` are not `ids` of the `report` model but of the model the report is working on
- this is a bugfix

I find it well to leave it like that until the base model is migrated

anthony-muschang:
@guewen :+1:

yvaucher:
agreed with @guewen for bug fix I prefer simple changes instead of trying to push other improvement changes such as API refactoring

pedrobaeza:
OK, let's keep it as is

PR #30, opened 2015-06-15
[FIX] test if the module is installed before monkey patching

vrenaville:
We need to test if the module is installed, if not it will raise an error if the code is present but not installed.


pedrobaeza:
:+1:

yvaucher:
:+1:

lmignon:
:+1: The errors in travis are unrelated to this change

guewen:
:+1:

PR #31, opened 2015-08-18
Add missing default oca icons

yvaucher:


pedrobaeza:
:+1:

lmignon:
:+1:  @yvaucher A nice to have should be to also fix the last remaining pep8 issues :smirk:

yvaucher:
@lmignon I fixed it in a separated PR: https://github.com/OCA/report-print-send/pull/32

yvaucher:
to merge carefully as modules have dependency on cups it could be another issue with static folders

yvaucher:
This is on branch 8.0, thus let's close it.

PR #32, opened 2015-08-18
[PEP8]

yvaucher:


base_report_to_printer/report.py
```diff
@@ -44,11 +44,12 @@ def print_document(self, cr, uid, ids, report_name, html=None,
             )
         return printer.print_document(report, document, report.report_type)
 
-    @api.v8
+    @api.v8  # noqa
```
pedrobaeza: Why do you need to add this?

yvaucher: Otherwise I get an flake8 error because of repetition of `print_document` and `get_pdf`

It might be my local config which is out of date. I'll try to remove it.

pedrobaeza: Yeah, I see. I didn't remember this one. Well, I think it's needed, because there's no other option in this case.

pedrobaeza: You can put in the comment the reason why it's used

yvaucher:
@pedrobaeza it is green without the noqa we certainly have deactivated that warning and my local config needs an update

pedrobaeza:
Cool then. Maybe it was addressed in the linter.

:+1:

guewen:
> @pedrobaeza it is green without the noqa we certainly have deactivated that warning and my local config needs an update

Yes that was what I was about to say, I'm sure we have explicitly deactivated this warning in the lint check.

guewen:
:+1:

lmignon:
:+1:  :clap: @yvaucher

PR #34, opened 2015-09-28
[IMP] base_report_to_printer : new setting

laetitia-gangloff:
per company and report

In multi-company environment, it can be usefull to define printer in function of the company instead of the user.
Use case is the following :
2 companies should print the same report on different printer. User can easily change company.

### DIFF
--- a/base_report_to_printer/ir_report.py
+++ b/base_report_to_printer/ir_report.py
@@ -85,8 +85,30 @@ class report_xml(orm.Model):
         'printing_action_ids': fields.one2many(
             'printing.report.xml.action', 'report_id', 'Actions',
             help='This field allows configuring action and printer on a per '
-                 'user basis'),
+                 'user / company basis'),
     }
+
+    def _get_company_id(self, cr, uid, context=None):
+        """
+        As it may happend that OpenERP force the uid to 1 to bypass rule (in
+        function field), we may sometimes use the company of user
+        id 1 instead of the good one. Because we found the real uid and
+        company_id in the context in that case, I return this one. It also
+        allow other module to give the proper company_id in the context.
+
+        If force_company is in context, use it in prioriy.
+
+        If company_id not in context, take the one from uid.
+        """
+        if context is None:
+            context = {}
+        res = context.get('force_company') or context.get('company_id')
+        if not res:
+            user_obj = self.pool['res.users']
+            res = user_obj.read(cr, uid, uid,
+                                ['company_id'],
+                                context=context)['company_id'][0]
+        return res
 
     def behaviour(self, cr, uid, ids, context=None):
         result = {}
@@ -99,6 +121,8 @@ class report_xml(orm.Model):
         if default_printer:
             default_printer = printer_obj.browse(cr, uid, default_printer,
                                                  context=context)
+
+        company_id = self._get_company_id(cr, uid, context)
 
         # Retrieve user default values
         user = self.pool['res.users'].browse(cr, uid, uid, context=context)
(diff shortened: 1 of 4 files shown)


base_report_to_printer/ir_report.py
```diff
@@ -85,9 +85,32 @@ def print_direct(self, cr, uid, report_id, result, format, printer,
...
     }
 
+    def _get_company_id(self, cr, uid, context=None):
+        """
+        As it may happend that OpenERP force the uid to 1 to bypass rule (in
+        function field), we may sometimes read the price of the company of user
+        id 1 instead of the good one. Because we found the real uid and
+        company_id in the context in that case, I return this one. It also
+        allow other module to give the proper company_id in the context (like
+        it's done in product_standard_margin for example).
```
guewen: Docstring comes obviously from elsewhere ;-)

laetitia-gangloff: Yes it is a copy (and not only for the docstring) from here: https://github.com/OCA/margin-analysis/blob/7.0/product_price_history/product_price_history.py#L224

Should I do something ?

guewen: Could you remove the incorrect mentions about prices?

base_report_to_printer/ir_report.py
```diff
@@ -118,6 +143,19 @@ def behaviour(self, cr, uid, ids, context=None):
...
+            if act_ids:
+                company_action = printing_act_obj.behaviour(
+                    cr, uid, act_ids[0], context)
+                action = company_action['action']
+                if company_action['printer']:
+                    printer = company_action['printer']
+
             # Retrieve report-user specific values
             act_ids = printing_act_obj.search(
                 cr, uid,
```
guewen, line 161: Since we have a company field now, maybe it could be nice to also look for a printer for a user in a company in the search domain below (for multi-company users).

I mean a query like the following (for : 

```
[('report_id', '=', report.id),
                  ('user_id', '=', uid),
                  ('action', '!=', 'user_default'),
                  '|', ('company_id', '=', company_id),
                   ('company_id', '=', False),
]
```

base_report_to_printer/printing_view.xml
```diff
@@ -107,7 +107,8 @@
       <field name="arch" type="xml">
         <form string="Report Printing Actions">
           <group col="2">
-              <field name="user_id"/>
+              <field name="company_id" groups="base.group_multi_company" attrs="{'required':[('user_id', '=', False)], 'invisible':[('user_id', '&lt;&gt;', False)]}"/>
+              <field name="user_id" attrs="{'required':[('company_id', '=', False)], 'invisible':[('company_id', '&lt;&gt;', False)]}"/>
```
guewen: Suggestion: replace `&lt;&gt;` by `!=`

base_report_to_printer/ir_report.py
guewen, line 161: I see that you added `attrs` to show only the company or the user field, which is working too, let me know what you think about my suggestion.

laetitia-gangloff, line 161: In my first implementation, I do this and I even add a printer configuration on company !
Then, I choose to keep this version with less change.

I think your suggestion give more configuration possibilities.

laetitia-gangloff:
@guewen I update following your suggestion.
So now, it is also possible to configure a report for the same user in different company.
Thank you for your review.

guewen:
Nice!
Thanks :+1:

yvaucher:
:+1:

PR #38, opened 2015-10-25
pass context to printer to use it for example in print_options()

intero-chz:
- context was lost while getting report
- now it will be passed using with_context
- could be used for print_options (example: pass copies amount for
  productlabals)

### DIFF
--- a/base_report_to_printer/report.py
+++ b/base_report_to_printer/report.py
@@ -42,7 +42,8 @@ class Report(models.Model):
             raise exceptions.Warning(
                 _('No printer configured to print this report.')
             )
-        return printer.print_document(report, document, report.report_type)
+        return printer.with_context(context).print_document(
+            report, document, report.report_type)
 
     @api.v8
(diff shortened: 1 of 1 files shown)


base_report_to_printer/report.py
```diff
@@ -42,7 +42,8 @@ def print_document(self, cr, uid, ids, report_name, html=None,
             raise exceptions.Warning(
                 _('No printer configured to print this report.')
             )
-        return printer.print_document(report, document, report.report_type)
+        return printer.with_context(context).print_document(
```
pedrobaeza, line 45: Shouldn't this be local_context?

intero-chz, line 45: I don't think 'must_skip_send_to_printer' option is necessary at this time anymore

guewen:
:+1:

pedrobaeza:
OK then :+1:

yvaucher:
:+1: merging

PR #40, opened 2015-11-17
[FIX][base_report_to_printer] active_ids without active_model

andrius-preimantas:
For some reason active_model is not set on pos.order. This sometimes results
inn exception because reports in some situations tries to browse active_ids
of unknown model. It actually should be fixed in core Odoo, however this fix
is quicker, easier to achieve.

Without this fix calling reporting printing from within POS front-end results in error.

Cheers
Andrius


yvaucher:
@andrius-preimantas as stated by @clonedagain could you make a bug report / PR on odoo and add it in the comment of this PR ?

PR #43, opened 2015-12-10
[IMP] Soft fail on missing cups import

RoelAdriaans:
cups is an external dependency, if it is not installed Odoo will not start.
OCA guidelines specify guidelines for External dependencies, code is from there.


guewen:
Thanks :+1:

sbidoul:
:+1:

pedrobaeza:
:+1:

PR #60, opened 2016-10-23
[9.0] Merge syleam printers module into base_report_to_printer

Garamotte:
I'm currently adding additional features of Syleam's [printers](https://github.com/syleam/printers) module into base_report_to_printer :
- [x] Add `printing.server` model : Management of multiple cups servers
- [x] Add `printing.job` model : Cups jobs tracking into Odoo
- [x] Add some methods on `printing.printer` model : `cancell_all_jobs`, `enable`, `disable`, etc.
- [x] Update cron jobs
- [x] Add migration script to create a `printing.server` record from configuration for existing databases

After this PR is merged, I plan to adapt our ZPL II labels printing module to work with base_report_to_printer.

### DIFF
--- a/base_report_to_printer/migrations/9.0.2.0.0/post-10-create_server_record.py
+++ b/base_report_to_printer/migrations/9.0.2.0.0/post-10-create_server_record.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 SYLEAM (<http://www.syleam.fr>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import SUPERUSER_ID, api
+from openerp.tools.config import config
+
+__name__ = 'Create a printing.server record from previous configuration'
+
+
+def migrate(cr, v):
+    with api.Environment.manage():
+        uid = SUPERUSER_ID
+        env = api.Environment(cr, uid, {})
+        env['printing.server'].create({
+            'name': config.get('cups_host', 'localhost'),
+            'address': config.get('cups_host', 'localhost'),
+            'port': config.get('cups_port', 631),
+        })
--- a/base_report_to_printer/models/printing_job.py
+++ b/base_report_to_printer/models/printing_job.py
@@ -1,0 +1,107 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 SYLEAM (<http://www.syleam.fr>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+from openerp import models, fields, api
+
+_logger = logging.getLogger(__name__)
+
+
+class PrintingJob(models.Model):
+    _name = 'printing.job'
+    _description = 'Printing Job'
+    _order = 'job_id_cups DESC'
+
+    name = fields.Char(help='Job name.')
+    active = fields.Boolean(
+        default=True, help='Unchecked if the job is purged from cups.')
+    job_id_cups = fields.Integer(
+        string='Job ID', required=True,
+        help='CUPS id for this job.')
+    server_id = fields.Many2one(
+        comodel_name='printing.server', string='Server',
+        related='printer_id.server_id', store=True,
+        help='Server which hosts this job.')
+    printer_id = fields.Many2one(
+        comodel_name='printing.printer', string='Printer', required=True,
+        ondelete='cascade', help='Printer used for this job.')
+    job_media_progress = fields.Integer(
+        string='Media Progress', required=True,
+        help='Percentage of progress for this job.')
+    time_at_creation = fields.Datetime(
+        required=True, help='Date and time of creation for this job.')
+    time_at_processing = fields.Datetime(
+        help='Date and time of process for this job.')
+    time_at_completed = fields.Datetime(
+        help='Date and time of completion for this job.')
+    job_state = fields.Selection(selection=[
+        ('pending', 'Pending'),
+        ('pending held', 'Pending Held'),
+        ('processing', 'Processing'),
+        ('processing stopped', 'Processing Stopped'),
+        ('canceled', 'Canceled'),
+        ('aborted', 'Aborted'),
+        ('completed', 'Completed'),
+        ('unknown', 'Unknown'),
+    ], string='State', help='Current state of the job.')
+    job_state_reason = fields.Selection(selection=[
+        ('none', 'No reason'),
+        ('aborted-by-system', 'Aborted by the system'),
+        ('compression-error', 'Error in the compressed data'),
+        ('document-access-error', 'The URI cannot be accessed'),
+        ('document-format-error', 'Error in the document'),
+        ('job-canceled-at-device', 'Cancelled at the device'),
+        ('job-canceled-by-operator', 'Cancelled by the printer operator'),
+        ('job-canceled-by-user', 'Cancelled by the user'),
+        ('job-completed-successfully', 'Completed successfully'),
+        ('job-completed-with-errors', 'Completed with some errors'),
+        ('job-completed(with-warnings', 'Completed with some warnings'),
+        ('job-data-insufficient', 'No data has been received'),
+        ('job-hold-until-specified', 'Currently held'),
+        ('job-incoming', 'Files are currently being received'),
+        ('job-interpreting', 'Currently being interpreted'),
+        ('job-outgoing', 'Currently being sent to the printer'),
+        ('job-printing', 'Currently printing'),
+        ('job-queued', 'Queued for printing'),
+        ('job-queued-for-marker', 'Printer needs ink/marker/toner'),
+        ('job-restartable', 'Can be restarted'),
+        ('job-transforming', 'Being transformed into a different format'),
+        ('printer-stopped', 'Printer is stopped'),
+        ('printer-stopped-partly',
+         'Printer state reason set to \'stopped-partly\''),
+        ('processing-to-stop-point',
+         'Cancelled, but printing already processed pages'),
+        ('queued-in-device', 'Queued at the output device'),
+        ('resources-are-not-ready',
+         'Resources not available to print the job'),
+        ('service-off-line', 'Held because the printer is offline'),
+        ('submission-interrupted', 'Files were not received in full'),
+        ('unsupported-compression', 'Compressed using an unknown algorithm'),
+        ('unsupported-document-format', 'Unsupported format'),
+    ], string='State Reason', help='Reason for the current job state.')
+
+    _sql_constraints = [
+        ('job_id_cups_unique', 'UNIQUE(job_id_cups, server_id)',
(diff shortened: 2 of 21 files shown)


base_report_to_printer/models/printing_printer.py
```diff
@@ -122,35 +75,47 @@ def print_document(self, report, content, format, copies=1):
...
-                    "you can reach it from the Odoo server.")
-                % (CUPS_HOST, CUPS_PORT))
-
-        options = self.print_options(report, format, copies)
+
+        connection = self.server_id._openConnection(raise_on_error=True)
+        options = self.print_options(
+            report=report, format=format, copies=copies)
+        # TODO : Check if this works in all cases
+        mime_type = 'application/octet-stream'
```
Garamotte: Last thing to check carefully : I didn't found any way to get the mime type of a variable's contents in python.

Does this type work for all file types and all printer drivers ?
If it doesn't, does someone know a way to find the mime type ?

Garamotte (commented):
This PR is ready to be reviewed.

There is a remaining question (inline comment), everything else should be OK.

Garamotte:
I removed the rewrite of the `print_document` method, because printing with writeRequestData needs the last fixes commited in pycups, which are not released yet (the current version has a bug with binary data). We'll have to wait for the pycups 1.9.74 release for this to work.
I kept the commit on a separate branch, to propose another PR when pycups' fix is released.

base_report_to_printer/models/printing_server.py
```diff
@@ -0,0 +1,233 @@
...
+        default=631, required=True, help='Port of the server.')
+    active = fields.Boolean(
+        default=True, help='If checked, this server is useable.')
+    printer_ids = fields.One2many(
+        comodel_name='printing.printer', inverse_name='server_id',
+        string='Printers List',
+        help='List of printers available on this server.')
+
+    @api.multi
+    def _openConnection(self, raise_on_error=False):
```
yvaucher: I understand that you used camel case as you open a `Connection` however you get the instance which is `connection` thus could we keep snake case here?

base_report_to_printer/models/printing_server.py
```diff
@@ -0,0 +1,233 @@
...
+
+            # Retrieve known uncompleted jobs data to update them
+            if which == 'not-completed':
+                min_jobs = job_obj.search([
+                    ('job_state', 'not in', (
+                        'canceled',
+                        'aborted',
+                        'completed',
+                    )),
+                    ('active', '=', True),
```
yvaucher: The active check is not necessary as applied by default

base_report_to_printer/models/printing_server.py
```diff
@@ -0,0 +1,233 @@
...
+                    'time-at-creation',
+                    'job-state',
+                    'job-state-reasons',
+                    'time-at-processing',
+                    'time-at-completed',
+                ])
+
+            # Retrieve known uncompleted jobs data to update them
+            if which == 'not-completed':
+                min_jobs = job_obj.search([
```
yvaucher: as you search with a `limit=1` variable should be singular.
Plus what does `min` means? is it to select the oldest unfinished job ?

base_report_to_printer/models/printing_server.py
```diff
@@ -0,0 +1,233 @@
...
+                    ('job_state', 'not in', (
+                        'canceled',
+                        'aborted',
+                        'completed',
+                    )),
+                    ('active', '=', True),
+                ], limit=1, order='jobid')
+                if min_jobs:
+                    min_job = min_jobs[0]
```
yvaucher: No need for that with a proper variable name and the `limit=1`

yvaucher (approved):
Nice workI! I only have few cosmetic code changes to propose but not blocking.

base_report_to_printer/models/printing_server.py
Garamotte: You're right, I'll change that.

base_report_to_printer/models/printing_server.py
Garamotte: Yes, `min` means that we search for the oldest : `order='jobid'`.
This is used in CUPS' `getJob` method arguments as a filter (`first_job_id=min_job.jobid`), there is no need to retrieve the full list, but only jobs not known as completed in Odoo's database.

base_report_to_printer/models/printing_server.py
Garamotte: Right, I don't remember why it was put here.

base_report_to_printer/models/printing_server.py
Garamotte: This is a remaining part of the old API, I'll fix that.

Garamotte:
@yvaucher Changes done, I renamed the `min_job` variable to `oldest_uncompleted_job`.

yvaucher:
@sylvain-garancher thanks

base_report_to_printer/migrations/9.0.2.0.0/post-10-create_server_record.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 SYLEAM (<http://www.syleam.fr>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import SUPERUSER_ID, api
+from openerp.tools.config import config
+
+__name__ = 'Create a pringing.server record from previous configuration'
```
lasley: s/pringing/printing

base_report_to_printer/models/printing_job.py
```diff
@@ -0,0 +1,106 @@
...
+
+class PrintingJob(models.Model):
+    _name = 'printing.job'
+    _description = 'Printing Job'
+    _order = 'jobid DESC'
+
+    name = fields.Char(help='Job name.')
+    active = fields.Boolean(
+        default=True, help='Unchecked if the job is purged from cups.')
+    jobid = fields.Integer(
```
lasley: Was there a reason for combining the words instead of snake case?

base_report_to_printer/models/printing_job.py
```diff
@@ -0,0 +1,106 @@
...
+    ], string='State Reason', help='Reason for the current job state.')
+
+    _sql_constraints = [
+        ('jobid_unique', 'UNIQUE(jobid, server_id)',
+         'The jobid of the printers job must be unique per server !'),
+    ]
+
+    @api.multi
+    def action_cancel(self):
+        return self.cancel()
```
lasley, line 92: action methods should `self.ensure_one`

base_report_to_printer/models/printing_printer.py
```diff
@@ -166,3 +121,41 @@ def set_default(self):
     @api.multi
     def get_default(self):
         return self.search([('default', '=', True)], limit=1)
+
+    @api.multi
+    def action_cancel_all_jobs(self):
```
lasley, line 159: action methods should `ensure_one`

base_report_to_printer/models/printing_server.py
```diff
@@ -0,0 +1,232 @@
...
+
+    @api.multi
+    def action_update_jobs(self):
+        if not self:
+            self = self.search([])
+        return self.update_jobs()
+
+    @api.multi
+    def update_jobs(self, which='all', first_job_id=-1):
+        job_obj = self.env['printing.job'].sudo()
```
lasley: Why was this not possible without sudo? I find this incredibly dangerous to blanket sudo an object for such a long method.

base_report_to_printer/models/printing_server.py
```diff
@@ -0,0 +1,232 @@
...
+                    'status': mapping.get(printer_info.get('printer-state'),
+                                          'unknown'),
+                    'status_message': printer_info.get(
+                        'printer-state-message', ''),
+                    'model': printer_info.get('printer-make-and-model', False),
+                    'location': printer_info.get('printer-location', False),
+                }
+                updated_printers.append(name)
+                if name not in existing_printers.keys():
+                    self.env['printing.printer'].sudo().create(printer_values)
```
lasley: Why sudo?

base_report_to_printer/models/printing_server.py
```diff
@@ -0,0 +1,232 @@
...
+                else:
+                    existing_printers[name].sudo().write(printer_values)
```
lasley: Why sudo?

lasley (changes requested):
Excellent additions @sylvain-garancher - some comments inline

base_report_to_printer/models/printing_job.py
Garamotte: I did that to differentiate CUPS' jobid (which is a regular integer in Odoo) from a real `*_id` many2one Odoo field.

base_report_to_printer/models/printing_server.py
Garamotte: I used sudo here because on our module, we run `update_jobs` (which runs `update_printers`) juste after sending a job to the printer, and not all users have the right to create a printer.

In this PR, I simply replaced the `update_printers` cron by `update_jobs`, which is run every minute, so I think the sudo is not needed anymore (unless we re-add the `update_jobs` call after each print, which doesn't seem useful).

base_report_to_printer/models/printing_server.py
Garamotte: Same reason as above, I think it can be removed.

base_report_to_printer/models/printing_server.py
Garamotte: Same reason as above, I think it can be removed.

Garamotte:
@lasley Typo and missing `ensure_one` fixed.
Waiting for answers to remove the `sudo` calls.

base_report_to_printer/models/printing_job.py
lasley: I had a feeling that was the case. Maybe `job_id_external`? Not a blocker or anything, it's just somewhat against PEP8

base_report_to_printer/models/printing_server.py
lasley: Ahh this makes sense. I agree that they don't seem needed & calling `update_jobs` every round seems excessive as well

Garamotte:
Sudo calls removed.
`jobid` field renamed to `job_id_cups`.

lasley (approved):
Thanks @sylvain-garancher 

👍  tentative on CI green

lasley:
Looks like we got a minor lint fail https://travis-ci.org/OCA/report-print-send/jobs/173318442#L432

Garamotte:
Oops, fixed :)

lasley (approved):
Woot 🍏

Garamotte:
Just re-added the `_prepare_update_from_cups` method, which is used in the `printer_tray` module.
My `update_printers` method (in `printing.server`) now calls this one, obviously.

Issue #61, opened 2016-10-24
Update Printers from CUPS throws error in ODOO 9.


metrogeekythinker:
```
Odoo Server Error
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 648, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 685, in dispatch
    result = self._call_function(*_self.params)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 321, in _call_function
    return checked_call(self.db, *args, *_kwargs)
  File "/usr/lib/python2.7/dist-packages/openerp/service/model.py", line 118, in wrapper
    return f(dbname, _args, *_kwargs)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 314, in checked_call
    result = self.endpoint(_a, *_kw)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 964, in __call__
    return self.method(_args, *_kw)
  File "/usr/lib/python2.7/dist-packages/openerp/http.py", line 514, in response_wrap
    response = f(_args, *_kw)
  File "/usr/lib/python2.7/dist-packages/openerp/addons/web/controllers/main.py", line 892, in call_button
    action = self._call_kw(model, method, args, {})
  File "/usr/lib/python2.7/dist-packages/openerp/addons/web/controllers/main.py", line 880, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, _args, *_kwargs)
  File "/usr/lib/python2.7/dist-packages/openerp/api.py", line 250, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/usr/lib/python2.7/dist-packages/openerp/api.py", line 353, in old_api
    recs = self.browse(cr, uid, [], context)
  File "/usr/lib/python2.7/dist-packages/openerp/api.py", line 250, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/usr/lib/python2.7/dist-packages/openerp/models.py", line 5380, in browse
    return self._browse(Environment(cr, uid, context or {}), ids)
  File "/usr/lib/python2.7/dist-packages/openerp/api.py", line 752, in __new__
    self.cr, self.uid, self.context = self.args = (cr, uid, frozendict(context))
TypeError: cannot convert dictionary update sequence element #0 to a sequence
```

[...]

PR #62, opened 2016-11-01
Set api.multi for action called as `object` on view

yvaucher:


yajo:
Why?

lasley:
@Yajo - action methods should all be `api.multi` per [Odoo docs](https://github.com/odoo/odoo/blob/731fe1c9abbcd3df0cf27053daaf3e7ca3d08ba3/doc/reference/guidelines.rst) - relevant line below because no headers to link:

> Action method : an object action method is prefix with action_. Its decorator is @api.multi, but since it use only one record, add self.ensure_one() at the beginning of the method.

We can't add an `ensure_one` here though, because it is also used as a [model method](https://github.com/OCA/report-print-send/search?q=action_ok&utf8=%E2%9C%93).

Realistically there should be two methods, but I won't block on that.

For v10, I would probably break the `api.multi` convention in order to use either [`model_cr`](https://github.com/odoo/odoo/blob/10.0/odoo/api.py#L369) or [`model_cr_context`](https://github.com/odoo/odoo/blob/10.0/odoo/api.py#L391), but we don't have those in v9.

Confirmed this fixes #61 - thanks @yvaucher

lasley (approved):

yajo:
Yes, I know, but in this case it does not matter, because the method does not use `self` for anything related to current record.

I mean, I won't block this, but that should make no difference.

From v10, the old api disappeared. Without decorator `@api.multi` will be implicit. `@api.model` makes not so much sense because you can still use `@api.multi` but call it with a zero-length recordset and it will be the same (just like this addon wants to do), so they will probably deprecate it along with `@api.one` in the future, as they said in last Odoo Experience.

Please fix that :x: in runbot at least.

yajo (approved):
Ah sorry I did not see that this aims to fix #61; then it's :+1: for me, just remember to fix the runbot's :x: before merging please :wink:

lasley:
I think Runbot is simply broken:

```
/srv/openerp/instances/openerp-oca-runbot/sandbox/bin/python: can't open file '/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3271346-yvaucher-patch-update-printers-8e5d6f/bin/openerp-server.py': [Errno 2] No such file or directory
```

I reran it yesterday with same results. Trying again now

Garamotte (approved):
:+1:
I saw that bug, and fixed it when doing my other PR #60.

But a small PR is much quicker to merge for a bugfix. I'll rebase my PR on 9.0 when this one is merged.

pedrobaeza:
OK, merging then

PR #66, opened 2016-11-17
[9.0] Add printer_zpl2 module

Garamotte:
A new module to print labels using a ZPL2 compatible printer (mostly Zebra printers, but other brands also use ZPL2).

The printer's configuration is altered (print width) before printing a label, then restored at the end of the label. This avoids issues when printing labels of a different width on the same printer (possibly from other applications).

### DIFF
--- a/printer_zpl2/README.rst
+++ b/printer_zpl2/README.rst
@@ -1,0 +1,99 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+ZPL II Label printing
+=====================
+
+This module extends the **Report to printer** (``base_report_to_printer``)
+module to add a ZPL II label printing feature.
+
+This module is meant to be used as a base for module development, and does not provide a GUI on its own.
+See below for more details.
+
+Installation
+============
+
+Nothing special, just install the module.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to *Settings > Printing > Labels > ZPL II*
+#. Create new labels
+
+It's also possible to add a label printing wizard on any model by creating a new *ir.values* record.
+For example, to add the printing wizard on the *product.product* model :
+
+.. code-block:: xml
+
+    <record model="ir.values" id="wizard_wizard_print_product_label">
+        <field name="name">Print Product Label</field>
+        <field name="key">action</field>
+        <field name="key2">client_action_multi</field>
+        <field name="model">product.product</field>
+        <field name="value" eval="'ir.actions.act_window,' + str(ref('printer_zpl2.action_wizard_print_record_label_view'))"/>
+    </record>
+
+Usage
+=====
+
+To print a label, you need to call use the label printing method from anywhere (other modules, server actions, etc.).
+
+.. code-block:: python
+
+    # Example : Print the label of a product
+    self.env['printing.label.zpl2'].browse(label_id).print_label(
+        self.env['printing.printer'].browse(printer_id),
+        self.env['product.product'].browse(product_id))
+
+You can also use the generic label printing wizard, if added on some models.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/144/9.0
+
+Known issues / Roadmap
+======================
+
+* Add a button to generate the ir.values for a model
+* Develop a "Designer" view in a separate module, to allow drawing labels with simple mouse clicks/drags
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/report-print-send/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Sylvain Garancher <sylvain.garancher@syleam.fr>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/printer_zpl2/models/printing_label_zpl2.py
+++ b/printer_zpl2/models/printing_label_zpl2.py
@@ -1,0 +1,188 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 SYLEAM (<http://www.syleam.fr>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import time
+import datetime
+import logging
+from openerp import api, exceptions, fields, models
+from openerp.tools.translate import _
+from openerp.tools.safe_eval import safe_eval
+
+_logger = logging.getLogger(__name__)
+
+try:
+    import zpl2
+except ImportError:
+    _logger.debug('Cannot `import zpl2`.')
+
+
+class PrintingLabelZpl2(models.Model):
+    _name = 'printing.label.zpl2'
+    _description = 'ZPL II Label'
+
+    name = fields.Char(required=True, help='Label Name.')
+    description = fields.Char(help='Long description for this label.')
+    model_id = fields.Many2one(
+        comodel_name='ir.model', string='Model', required=True,
+        help='Model used to print this label.')
+    origin_x = fields.Integer(
+        required=True, default=10,
+        help='Origin point of the contents in the label, X coordinate.')
+    origin_y = fields.Integer(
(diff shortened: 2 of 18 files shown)


Garamotte:
@pedrobaeza @lasley Can you tag this PR as "Needs review" ?
Everything is OK on my side, tested on several Zebra printers without any issue.

pedrobaeza:
Done

printer_zpl2/README.rst
```diff
@@ -0,0 +1,90 @@
...
+   :alt: License: AGPL-3
+
+=====================
+ZPL II Label printing
+=====================
+
+This module extends the **Report to printer** (``base_report_to_printer``)
+module to add a ZPL II label printing feature.
+
+It's done to be used from other modules, or from python code blocks.
```
lasley: I'm not quite sure what you're going for with this statement, so I can't make any recommendations, but it's unclear IMO. Is it that this is a base module, meant to be used with other modules? That's what I'm thinking from reading the rest of ReadMe?

printer_zpl2/lib/zpl2.py
```diff
@@ -0,0 +1,335 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 SYLEAM (<http://www.syleam.fr>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+class Zpl2(object):
```
lasley: Any reason why you didn't create a python lib out of this? IMO this would totally be useful out of Odoo context.

printer_zpl2/lib/zpl2.py
```diff
@@ -0,0 +1,335 @@
...
+
+    def output(self):
+        """ Return the full contents to send to the printer """
+        return self._buffer
+
+    def _enforce(self, value, minimum=1, maximum=32000):
+        """ Returns the value, forced between minimum and maximum """
+        return min(max(minimum, value), maximum)
+
+    def _writeCommand(self, data):
```
lasley: Why not PEP-8 method names?

printer_zpl2/lib/zpl2.py
```diff
@@ -0,0 +1,335 @@
...
+
+        # Add all arguments in the list, if they exist
+        for argument in arguments:
+            if kwargs.get(argument, None) is not None:
+                if isinstance(kwargs[argument], bool):
+                    kwargs[argument] = kwargs[argument] and 'Y' or 'N'
+                commandArguments.append(kwargs[argument])
+
+        # Return a zebra formatted string, with a comma between each argument
+        return ','.join(map(str, commandArguments))
```
lasley: Nice trick, I like

printer_zpl2/lib/zpl2.py
```diff
@@ -0,0 +1,335 @@
...
+        """ Add the data stop command to the buffer """
+        return '^FS'
+
+    def _fieldData(self, data):
+        """ Add data to the buffer, between start and stop commands """
+        command = ''
+        command += self._fieldDataStart()
+        command += data
+        command += self._fieldDataStop()
+        return command
```
lasley: IMO this would be cleaner:
```
return '%s%s%s' % (
    self.self._fieldDataStart(), data, self._fieldDataStop()
)
# OR 
''.join((
    self.self._fieldDataStart(), data, self._fieldDataStop()
))
```

printer_zpl2/lib/zpl2.py
```diff
@@ -0,0 +1,335 @@
...
+
+    def fontData(self, right, down, fieldFormat, data):
+        """ Add a full text in the buffer, with needed formatting commands """
+        command = ''
+        command += self._fieldOrigin(right, down)
```
lasley: This isn't very efficient. Take a look at this awesome comparison of string concats - https://waymoot.org/home/python_string/

Strings are immutable, so each time you `+=` you're actually creating three strings in memory then discarding two. (created `old` string, `add` string, `concatenated` string - discarded `old` and `add`)

printer_zpl2/lib/zpl2.py
```diff
@@ -0,0 +1,335 @@
...
+                if kwargs['interpretation_line'] and not kwargs[
+                   'interpretation_line_above']:
+                    kwargs['interpretation_line'] = 'B'
+                else:
+                    kwargs['interpretation_line'] = 'A'
+            return '4' + self._generateArguments(arguments, kwargs)
+
+        def _pdf417(**kwargs):
+            arguments = [
+                'orientation',
```
lasley: IMO these strings should all be constants. Definitely a stylistic recommendation so feel free to ignore, I've just always found it incredibly useful to have important strings statically defined - particularly if used more than once.

lasley (changes requested):
❤️

lasley:
Oh also please bump the version of `printer_tray` up a patch version to indicate the slight change made there.

printer_zpl2/README.rst
Garamotte: With this statement, I want to explain that the module mainly provides the way to configure labels, a method to print a label (to be called the way you want) and a generic wizard that can be added on any model by a custom module. It doesn't have a full GUI with lots of options.
Maybe my english is not good enough here :D

printer_zpl2/lib/zpl2.py
Garamotte: The class has been developped with the module since a long time, but as it is really generic, creating a separate python lib is a good idea.
I'll do that :)

lasley: Ping me when you do and I'll be happy to review it. I recommend maybe also taking a look at a [helper library I've been playing with that's fairly similar to MQT, but for py-libs](https://github.com/LasLabs/python-quality-tools).

printer_zpl2/lib/zpl2.py
Garamotte: I don't remember, that's a very old name :)
I'll review the method names.

printer_zpl2/lib/zpl2.py
Garamotte: These correspond to Odoo fields names, so I think this can (should ?) be constants.
I'll refactor that when moving the lib to a separate python package.

printer_zpl2/lib/zpl2.py
Garamotte: I prefer using str.format, but you're right, this has to be refactored.

printer_zpl2/lib/zpl2.py
Garamotte: Totally right... I don't know why I did that this dumb way...

printer_zpl2/README.rst
lasley: Awesome thanks for the clarification! What about something like:

> This module is meant to be used as a base for module development, and does not provide a GUI on its own. See the configuration section for more details.

Does that nail the intent?

Garamotte: Exactly, thanks. I'll take your statement.
I think the readme can also be improved for that, as the `_print_label` method is not mentionned.

lasley: Oh yeah definitely a good call mentioning that in the ReadMe too seeing as it has the error in there by model 👍

Garamotte:
@lasley First step done : Refactoring to take all your comments in account (pep8 names, constants, etc.). I'll separate the zpl2 python lib when the big changes are finished.

lasley:
@sylvain-garancher - I agree that completing the review here is a good call

@tedsalmon please review

printer_zpl2/models/printing_label_zpl2_component.py
```diff
@@ -0,0 +1,137 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 SYLEAM (<http://www.syleam.fr>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from ..lib import zpl2
+from openerp import models, fields
```
tedsalmon: Imports should be alphabetized.

printer_zpl2/models/printing_label_zpl2.py
```diff
@@ -0,0 +1,181 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 SYLEAM (<http://www.syleam.fr>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import time
+import datetime
+from ..lib import zpl2
+from openerp import models, api, fields, exceptions
```
tedsalmon: Imports should be alphabetized

printer_zpl2/wizard/print_record_label.xml
```diff
@@ -0,0 +1,37 @@
...
+        <field name="type">ir.actions.act_window</field>
+        <field name="res_model">wizard.print.record.label</field>
+        <field name="view_type">form</field>
+        <field name="view_mode">form</field>
+        <field name="domain">[]</field>
+        <field name="context">{}</field>
+        <field name="target">new</field>
+    </record>
+
+    <!--
```
tedsalmon: Any reason this is here too? Seems a little redundant, given this is in the README. :)

tedsalmon (approved):
LGTM minus some minor nit-picks

printer_zpl2/wizard/print_record_label.xml
Garamotte: This was here before I create the README, and I forgot to remove :D

Garamotte:
Comments from @tedsalmon fixed.

I've split the `zpl2` lib into a separate python module.
The git repository is here : https://github.com/syleam/python-zpl2
This lib is pip-installable with `pip install zpl2`.

lasley (approved):

sla-subteno-it:
👍

lasley:
@sylvain-garancher do you plan on migrating this to v10?

sla-subteno-it:
@lasley I just migrated in v10. I sent a new PR.

PR #67, opened 2016-11-20
[FIX] 9.0 Issue#63 printing now working

auphofBSF:
missing 1st parameter (rec_id) on get_pdf method

Issue #63, opened 2016-11-02
error printing invoice from POS


efarias:
Trying to print an invoice from POS, odoo 9.0, it gives me this error:

> 2016-11-02 21:09:32,939 15700 ERROR BaltoluProductionStable2016 openerp.http: Exception during JSON request handling.
> Traceback (most recent call last):
>   File "/opt/odoo/openerp/http.py", line 643, in _handle_exception
>     return super(JsonRequest, self)._handle_exception(exception)
>   File "/opt/odoo/openerp/http.py", line 680, in dispatch
>     result = self._call_function(**self.params)
>   File "/opt/odoo/openerp/http.py", line 316, in _call_function
>     return checked_call(self.db, *args, **kwargs)
>   File "/opt/odoo/openerp/service/model.py", line 118, in wrapper
>     return f(dbname, *args, **kwargs)
>   File "/opt/odoo/openerp/http.py", line 309, in checked_call
>     result = self.endpoint(*a, **kw)
>   File "/opt/odoo/openerp/http.py", line 959, in __call__
>     return self.method(*args, **kw)
>   File "/opt/odoo/openerp/http.py", line 509, in response_wrap
>     response = f(*args, **kw)
>   File "/opt/odoo/addons/web/controllers/main.py", line 892, in call_kw
>     return self._call_kw(model, method, args, kwargs)
>   File "/opt/odoo/addons/web/controllers/main.py", line 884, in _call_kw
>     return getattr(request.registry.get(model), method)(request.cr, request.uid, *args, **kwargs)
>   File "/opt/odoo/openerp/api.py", line 238, in wrapper
>     return old_api(self, *args, **kwargs)
>   File "/opt/odoo/openerp/api.py", line 369, in old_api
>     result = method(recs, *args, **kwargs)
>   File "/opt/odoo/addons/base_report_to_printer/models/report.py", line 21, in print_document
>     document = rec_id.get_pdf(report_name, html=html, data=data)
>   File "/opt/odoo/openerp/api.py", line 236, in wrapper
>     return new_api(self, *args, **kwargs)
> TypeError: get_pdf() takes at least 3 arguments (4 given)

auphofBSF:
Same Issue as **efarias**, I have looked into it and found a fix,  as per image,   rec_id requires to be passed to the get_pdf call as 1st parameter
![image](https://cloud.githubusercontent.com/assets/11743864/20461585/795b8830-af67-11e6-9c1b-c913613637a5.png)

[...]

PR #68, opened 2016-11-24
[10.0][MIG] base_report_to_printer

Garamotte:
Migration of base_report_to_printer to v10.0
This migration includes improvements from #60

The deprecated methods in `printing.printer` have been removed (methode to update printers from cups, that is now done in `printing.server`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ python:
 
 addons:
   apt:
+    sources:
+      - pov-wkhtmltopdf
     packages:
       - expect-dev  # provides unbuffer utility
       - python-lxml  # because pip installation is slow
@@ -18,6 +20,11 @@ addons:
       - python-yaml
       - cups
       - libcups2-dev
+      - wkhtmltopdf  # only add if needed and check the before_install section below
+
+before_install:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
 
 env:
   global:
--- a/base_report_to_printer/README.rst
+++ b/base_report_to_printer/README.rst
@@ -90,6 +90,7 @@ Contributors
 * Lionel Sausin <ls@numerigraphe.com>
 * Guewen Baconnier <guewen.baconnier@camptocamp.com>
 * Dave Lasley <dave@laslabs.com>
+* Sylvain Garancher <sylvain.garancher@syleam.fr>
 
 Maintainer
 ----------
--- a/base_report_to_printer/__manifest__.py
+++ b/base_report_to_printer/__manifest__.py
@@ -8,7 +8,7 @@
 
 {
     'name': "Report to printer",
-    'version': '9.0.2.0.0',
+    'version': '10.0.1.0.0',
     'category': 'Generic Modules/Base',
     'author': "Agile Business Group & Domsense, Pegueroles SCP, NaN,"
               " LasLabs, Odoo Community Association (OCA)",
@@ -27,7 +27,7 @@
         'views/ir_actions_report_xml_view.xml',
         'wizards/printing_printer_update_wizard_view.xml',
     ],
-    'installable': False,
+    'installable': True,
     'application': True,
     'external_dependencies': {
         'python': ['cups'],
--- a/base_report_to_printer/__manifest__.py
+++ b/base_report_to_printer/__manifest__.py
@@ -8,7 +8,7 @@
 
 {
     'name': "Report to printer",
-    'version': '9.0.1.0.0',
+    'version': '9.0.2.0.0',
     'category': 'Generic Modules/Base',
     'author': "Agile Business Group & Domsense, Pegueroles SCP, NaN,"
               " LasLabs, Odoo Community Association (OCA)",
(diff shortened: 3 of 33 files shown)


Garamotte:
@pedrobaeza @gurneyalex 
Do you know why Travis build crashes ? It seems that the PDF returned by wkhtmltopdf is invalid.
I tried other ways of installing wkhtmltopdf, but it didn't work.

.travis.yml
```diff
@@ -18,6 +20,11 @@ addons:
       - python-yaml
       - cups
       - libcups2-dev
+      - wkhtmltopdf  # only add if needed and check the before_install section below
+
+before_install:
+  - "export DISPLAY=:910.0"
```
pedrobaeza: In the template we have `99'  instead of '910`

Garamotte:
Ok, thanks, everything is OK now :)
I picked the value in another repo, and after a quick grep, I found 7 repos using 910 and only one using 99. Maybe wkhtmltopdf is not used in tests on these repos.

lasley (approved):
Code LGTM thanks @sylvain-garancher 

I'm installing this on a prod, so I'll report back if anything explodes. If I don't report back, we can assume this is also a functional 👍

Garamotte:
I still have to split some tests, and edit some comments to follow guidelines, as you stated on my other PR a month ago, but this cannot break the module :)

lasley:
Sweet thanks for the update @sylvain-garancher, and for making sure all your modules are top notch before merge 🚀 

So far our few tests on the prod box worked well, so functionally we're good in that respect too it seems.

lasley:
@sylvain-garancher - FYI please make sure to include #73

tedsalmon:
@lasley,

I have submitted a PR (https://github.com/syleam/report-print-send/pull/1) to forward-port these changes into @sylvain-garancher branch :)

Garamotte:
Thanks, PR merged.

Garamotte:
I'll include https://github.com/OCA/report-print-send/pull/74 when it's merged in 9.0.
Everything else OK on my side.

lasley:
Long term test on my end is 👍  as well. #74 has been merged

Garamotte:
I received a PR on the v10 migration branch : https://github.com/syleam/report-print-send/pull/2
Looking the issues list, I also found two issues and a PR for this bug on v9 : https://github.com/OCA/report-print-send/issues/75 / https://github.com/OCA/report-print-send/issues/63 / https://github.com/OCA/report-print-send/pull/67

I think the best would be to fix the bug in v9 (with a unit test, if possible, but I don't have time actually), then I'll cherry-pick the fix into my v10 migration branch.

Garamotte:
@lasley Bug fix merged into 9.0 and cherry-picked here.
You can replace the "WIP" tag by "Needs review".

lasley:
@sylvain-garancher Done! I'll also toss this tab in my to-review pile & come back to it soon 😄

lasley (approved):

lasley:
@tedsalmon please review

peterfbrandl:
hope I am right here:

after installing the module it is impossible to print a report in "HTML" directly to the screen. The whole system seems to be in an endless loop. In the background you can see the result of the html-report but the loading symbol is still there. But this is only after installing the report-to-printer module

Thank you for your help

screen_report

lasley:
@peterfbrandl - thanks for the report. I can confirm this issue exists on Runbot:
* Install `sale` app
* Set quotation report type to HTML
* Navigate to a quote and print:
* HTML report is generated and shown, but the UI is never released. No error in console:
  * 
image


The error clears itself after uninstalling the module (the infinite loop will need to time out first, as well).

Garamotte:
Thanks for giving a simple way to reproduce the issue.
I won't have time to look at this issue in the next days, any help is welcome :)

I think this is caused by the JS part : https://github.com/OCA/report-print-send/blob/4067f73e19e3238f13830b9e4d67b4af1d49f993/base_report_to_printer/static/src/js/qweb_action_manager.js
We see that `blockUI` is called in the beginning of the method, then the super is called because we don't use qweb-pdf here. Maybe the `unblockUI` method was called in the standard v9, and is no longer called on v10 ?

Garamotte:
@peterfbrandl @lasley Fixed.
The UI is blocked (and then, unblocked) only when using qweb-pdf reports in v10, while it was blocked/unblocked for all reports on v9.
I simply moved the UI block call in the qweb-pdf case.

lasley (approved):
Confirmed fix, thanks @sylvain-garancher!

peterfbrandl:
Confirmed fix. Thank you very much @sylvain-garancher

lasley:
@peterfbrandl - Please use the Github review feature (in the Files Changed tab) so that we may count your review for merge.

lasley:
I'm merging - we don't have the amount of reviews we need, but this has been in production for a while & I infer that @peterfbrandl is also using this as well - even though no explicit review was provided. 

Furthermore, we now have other contributors trying to upgrade this as well (#81). If something is broken & gets merged during this, I will take full responsibility and fix.

PR #70, opened 2016-12-28
[9.0][IMP] Improove base_report_to_printer security usabilty

jjscarafia:
1. Add category for grupos Printing for "Print User" and "Print Manager" so that it is rendered like a selection on groups set up (check images)
2. Make res.groups and ir.model.access loaded by this module "updatable" as it is on odoo standards

Before:
seleccion_020

After:
seleccion_019

### DIFF
--- a/base_report_to_printer/migrations/9.0.2.1.0/pre-migration.py
+++ b/base_report_to_printer/migrations/9.0.2.1.0/pre-migration.py
@@ -1,24 +1,6 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenUpgrade module for Odoo
-#    @copyright 2015-Today: Odoo Community Association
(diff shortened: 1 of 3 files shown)


base_report_to_printer/migrations/9.0.2.1.0/pre-migration.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Use short headers

pedrobaeza (approved):

jjscarafia:
@pedrobaeza Done, Thanks!

gonzaloruzafa (approved):
Functional test Ok

moylop260 (approved):

pedrobaeza (approved):

PR #71, opened 2016-12-28
[9.0][MIG] Improove migration from v8 of base_report_to_printer

jjscarafia:
Fix migration from v8. xml id for some none updateble groups and acces rights where renamed as you can see here: [V8](https://github.com/OCA/report-print-send/blob/8.0/base_report_to_printer/security/security.xml) vs [V9](https://github.com/OCA/report-print-send/blob/9.0/base_report_to_printer/security/security.xml)

### DIFF
--- a/base_report_to_printer/migrations/9.0.1.0.0/pre-migration.py
+++ b/base_report_to_printer/migrations/9.0.1.0.0/pre-migration.py
@@ -1,23 +1,6 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenUpgrade module for Odoo
-#    @copyright 2014-Today: Odoo Community Association
-#
-#    This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


gonzaloruzafa (approved):
Everything looks ok

base_report_to_printer/migrations/9.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Short header

jjscarafia: jejej, sorry, I've just see the other comment for the same on RMA repository. Done

pedrobaeza (approved):

moylop260 (approved):
Thanks

PR #72, opened 2016-12-28
[9.0][FIX] FIX Migration script

jjscarafia:
Fix https://github.com/OCA/report-print-send/pull/71
Sorry @pedrobaeza @moylop260 @gonzaloruzafa. I set one wrong id repacement for group manager (is [thisone](https://github.com/OCA/report-print-send/blob/9.0/base_report_to_printer/security/security.xml#L4) and not [thisone](https://github.com/OCA/report-print-send/blob/9.0/base_report_to_printer/security/security.xml#L11). Can you check this fix please?


gonzaloruzafa (approved):
+1

moylop260 (approved):
oh yeah xml_ids similar names

pedrobaeza (approved):
No problem

PR #74, opened 2017-01-17
[FIX] Fixed the res.users view

Garamotte:
The string attribute should not be used as a selector, because it is translatable.

cc @RoelAdriaans-B-informed

### DIFF
--- a/base_report_to_printer/views/res_users_view.xml
+++ b/base_report_to_printer/views/res_users_view.xml
@@ -6,12 +6,12 @@
     <field name="model">res.users</field>
     <field name="inherit_id" ref="base.view_users_form" />
     <field name="arch" type="xml">
-      <page string="Preferences">
-        <group name="printing">
+      <xpath expr="//group[@name='preferences']/ancestor::page" position="inside">
+        <group string="Printing" name="printing">
           <field name="printing_action"/>
           <field name="printing_printer_id"/>
         </group>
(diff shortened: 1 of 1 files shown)


RoelAdriaans (approved):
Checked in Odoo v9 Enterprise with other translations, fixes issue I was having

base_report_to_printer/views/res_users_view.xml
```diff
@@ -6,12 +6,12 @@
     <field name="model">res.users</field>
     <field name="inherit_id" ref="base.view_users_form" />
     <field name="arch" type="xml">
-      <page string="Preferences">
-        <group name="printing">
+      <xpath expr="//page[group/group[@name='preferences']]" position="inside">
```
pedrobaeza: You can put it even more generic:
```
<xpath expr="//group[@name='preferences']/ancestor::page" position="inside">

Garamotte: Thanks, I didn't know this syntax, I update with your version.

If there are several `page` ancestors, would it take the first or the last ?

pedrobaeza: It would take the first, although you can put also indexes if needed, but it's not the usual case.

pedrobaeza (approved):

lasley (approved):

PR #76, opened 2017-02-22
[MIG] migration to odoo 10

JosefKaser:
Changes to use the modul with odoo 10.


pedrobaeza:
There's already a migration in https://github.com/OCA/report-print-send/pull/68. Please check before duplicating efforts.

Issue #75, opened 2017-02-09
server print error and fails to link CUPS when printing 


nonofpy:
Hi All, 

I am working in Odoo 9.0 and installed the module 'report to printer'. But I got the following issues.
CUPS connects to the Odoo successfully. Because I can get the server and printers information in CUPS as well as the real-time jobs. 

1. server error when I print  as 'pdf' type
- configure the reports settings
![image](https://cloud.githubusercontent.com/assets/20656150/22767131/7c7f866a-eeb4-11e6-8142-d4e3bb1ad660.png)
- print the SO, and get the below error 
![image](https://cloud.githubusercontent.com/assets/20656150/22767155/96f3238a-eeb4-11e6-81fd-afb96f568f34.png)

2. still link to the local printer instead of CUPS
- configure the reports -> as 'html' (without server errors)
- result -> popup a local printing page

Do I miss something in the configuration ? It seems that Odoo can not link to CUPS.

Please help!

Thank you so much. 

Regards,
Betty

Garamotte:
Fixed by https://github.com/OCA/report-print-send/pull/77

PR #77, opened 2017-02-23
[FIX] Fix the new API migration of base_report_to_printer

Garamotte:
The new API migration was made to `@api.multi` because of the `cr, uid, ids` signature, but `ids` was the ids of the records to print here, not the report's ids. This method should be `@api.model`.
Also, the new API version of `get_pdf` get directly the ids of the records to print in the standard module, not a recordset.

Fixes https://github.com/OCA/report-print-send/issues/63 and https://github.com/OCA/report-print-send/issues/75
Supersedes https://github.com/OCA/report-print-send/pull/67

### DIFF
--- a/base_report_to_printer/models/report.py
+++ b/base_report_to_printer/models/report.py
@@ -8,28 +8,19 @@ from openerp import models, exceptions, _, api
 class Report(models.Model):
     _inherit = 'report'
 
-    @api.multi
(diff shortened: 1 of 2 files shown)


lasley (approved):
Good catch @sylvain-garancher, thanks

yvaucher (approved):

bealdav:
:+1:

Issue #56, opened 2016-08-16
print an invoice each time you send it by mail


PhilippeBach:
I use to print my invoices when they are created but i use the email to customer fonctionality to send the invoice by mail to my customer and odoo send a pdf to the printer each time i create a mail, ever when i change the email template 
it use a lot of paper 

on Odoo V8

lasley:
Hi @PhilippeBach - thanks for the report. 

To confirm - when you send an invoice using the `Send by Email` button, this is both sending an email to the customer and printing via the external interface?

Does this happen on all users?

PhilippeBach:
yes, on all users we use to print invoices directly to printer for all users

i think it is because the invoice created in email dont have the same name, so it can't detect the the invoice was created before .
if we change the template with an other pdf name  it print it also

Garamotte:
Someone else reported me the issue, while testing the v10 migration.

Looking at te code, I think the fix would be to inherit the `generate_email` method of `mail.template` to add `must_skip_send_to_printer` in the context.

Garamotte:
Fixed by https://github.com/OCA/report-print-send/pull/78
@PhilippeBach You'll just have to install the new `base_report_to_printer_mail` module, if it doesn't install automatically.

PR #78, opened 2017-03-25
[ADD] Add the base_report_to_printer_mail module

Garamotte:
An auto-install module for compatibility between base_report_to_printer and mail by avoiding to print reports when generating them for email attachments.

Fixes https://github.com/OCA/report-print-send/issues/56

### DIFF
--- a/base_report_to_printer_mail/README.rst
+++ b/base_report_to_printer_mail/README.rst
@@ -1,0 +1,53 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+========================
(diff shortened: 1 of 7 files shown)


pedrobaeza:
As mail is needed for everything, why don't you directly include this in main module and depend on mail, so you avoid glue modules?

Garamotte:
That's also an option, but creating a database with only `base_report_to_printer` doesn't install `mail`, so I decided to create a separate module, even though I don't know any Odoo production database without the mail module installed.

If you prefer to add a dependency, I can do that instead of the small glue module :)

pedrobaeza (approved):
OK, let's go this way to have a proper modularity, although I also don't know any installation without mail module installed, as it's a dependency of almost everything.

lasley (approved):
Thanks @sylvain-garancher

PR #82, opened 2017-04-06
[8.0][FIX] Move dependencies from Travis file

lasley:
This moves the depends to a requirements and oca_depends file, instead of Travis.


pedrobaeza (approved):

pedrobaeza:
This can be fast-track as the CIs result is green

PR #83, opened 2017-04-06
[9.0][FIX] Move dependencies from Travis file

lasley:
This moves the depends to a requirements and oca_depends file, instead of Travis.


pedrobaeza (approved):
This can be fast-track also

PR #84, opened 2017-04-11
[10] FIX format parameter

angelmoya:
To be able to print files in raw format it's needed to set this parameter.

Previous code was not able to print raw files, but it didn't get error because 'format' used on line 107 is assumed as a format funtion, because format parameter was not on this scope.

### DIFF
--- a/base_report_to_printer/models/printing_printer.py
+++ b/base_report_to_printer/models/printing_printer.py
@@ -95,10 +95,10 @@ class PrintingPrinter(models.Model):
         finally:
             os.close(fd)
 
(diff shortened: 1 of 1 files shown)


angelmoya:
@lasley @alnslang maybe you want to take look this.

This fix is needed for new module that I will push later, called base_delivery_carrier_label_direct_print, it will send raw file in zpl format to printer, I'm testing and it's working right.

elicoidal (approved):

lasley (approved):
Thanks @angelmoya - LGTM

lasley:
Fast tracking - this is fixing a break, and is obvious

PR #86, opened 2017-05-01
Update README.rst

blutecsolutions:
Provided steps to installing PyCups

### DIFF
--- a/base_report_to_printer/README.rst
+++ b/base_report_to_printer/README.rst
@@ -29,13 +29,12 @@ Installation
 
 * Install PyCups - https://pypi.python.org/pypi/pycups
 
-1) sudo apt-get install cups
+.. code-block:: bash
 
-2) sudo apt-get install libcups2-dev
-
-3) sudo apt-get install python-dev OR sudo apt-get install python3-dev
-
-4) sudo easy_install pycups OR sudo pip install pycups
+   sudo apt-get install cups
+   sudo apt-get install libcups2-dev
+   sudo apt-get install python-dev OR sudo apt-get install python3-dev
+   sudo easy_install pycups OR sudo pip install pycups
 
 
 Configuration
--- a/base_report_to_printer/README.rst
+++ b/base_report_to_printer/README.rst
@@ -29,6 +29,14 @@ Installation
 
 * Install PyCups - https://pypi.python.org/pypi/pycups
 
+1) sudo apt-get install cups
+
+2) sudo apt-get install libcups2-dev
+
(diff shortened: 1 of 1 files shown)


base_report_to_printer/README.rst
```diff
@@ -29,6 +29,14 @@ Installation
 
 * Install PyCups - https://pypi.python.org/pypi/pycups
 
+1) sudo apt-get install cups
```
elicoidal: Thanks for your proposal. 

I personally would rather do something like this: 
```
.. code-block:: bash

   sudo apt-get install cups
   sudo apt-get install libcups2-dev
   sudo apt-get install python-dev OR sudo apt-get install python3-dev
   sudo easy_install pycups OR sudo pip install pycups
```

blutecsolutions: Thanks for the suggestion: Please update it :)

thomich:
thats a very important README change, please check and merge...

pedrobaeza:
But it's very platform/version dependent. If someone doesn't use Ubuntu or Ubuntu 18.04 uses other names, or Odoo switches to Python 3.... I'm not in favor of these kind of instructions.

thomich:
then create platform dependent README's.. 

if we we always think, that we can't create a proper README cause it's platform/version dependent, then we never can make good README's

yvaucher:
It's better to have a install procedure for one platform instead of none

:+1:

yvaucher (approved):

pedrobaeza:
Well, you shouldn't merge this without consent, @yvaucher 

If you merge it, at least put that the instruction are for Ubuntu 14.04-16.04 or similar.

yvaucher:
@pedrobaeza This is doc it has no impact on the module itself. And we strongly need doc for our modules. Thus I prefer to do optimistic merge in that case.

This PR is 22 days old and has 2 approvals with @thomich and mine.

pedrobaeza:
The problem is with my disapproval, @yvaucher. You can have 200 approvals but if someone expresses any concern, you should wait until it's solved. I'm not saying of not merging, but not merging as is.

yvaucher:
@pedrobaeza Your disapproval was not only a comment but it also didn't provide any guidance on how to make you happy. Plus your disapproval was disapproved :)

This PR has been there for 22 days the author will probably not read it soon and noone wanted to make it nicer. So this is one step in the good direction. Better than nothing.

pedrobaeza:
OK, I don't want to discuss more, but this is not the good way to do things. Just wait a bit. You know that I answer very fast. Anyway, you can add the sentence about the SO for completing the PR.

PR #89, opened 2017-06-07
[MIG] Migrated base_report_to_printer_mail to 8.0

robch2342:


base_report_to_printer_mail/__openerp__.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
Garamotte, line 2: Please use the short header, like in the template : https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py

base_report_to_printer_mail/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+
+{
+    'name': 'Report to printer - Mail',
+    'version': '8.0.1.0.0',
+    'category': 'Reporting',
+    'summary': 'Avoid printing a report on email send',
+    'description': """
+
+""",
+    'images': [],
```
Garamotte, line 30: Remove empty keys.

base_report_to_printer_mail/email_template.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
Garamotte, line 2: Use the short header.

base_report_to_printer_mail/email_template.py
```diff
@@ -0,0 +1,33 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import osv
+
+import logging
+_logger = logging.getLogger(__name__)
+
+class email_template(osv.osv):
```
Garamotte, line 27: Prefer using `models.Model` instead of `osv.osv` since v8.0.

base_report_to_printer_mail/email_template.py
```diff
@@ -0,0 +1,33 @@
...
+    _inherit = "email.template"
+    
+    def generate_email_batch(self, cr, uid, template_id, res_ids, context={}, fields=None):
```
Garamotte, line 30: Don't use `context={}` but `context=None` instead.
More information about that:
- http://python-guide-pt-br.readthedocs.io/en/latest/writing/gotchas/#mutable-default-arguments
- https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html#using-a-mutable-default-value-as-an-argument

base_report_to_printer_mail/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+    'author': 'Vertel AB',
+    'website': 'http://www.vertel.se, Odoo Community Association (OCA)',
```
Garamotte, line 32: OCA should be in the `author` key instead of `website`.

Garamotte (changes requested):
Thanks for the backport.
Is it possible to backport the unit test too ?

pedrobaeza:
And you should preserve commit history. Use https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0 switching the branches names.

PR #90, opened 2017-06-08
Add a button to generate the ir.values for a model

JosDeGraeve:
On the "ZPL II Labels" form, there are buttons added "Add in the 'More' menu" and
"Remove from the 'More' menu".

This way a user can easily set a print function for his label.

To add the buttons, i had to add a <sheet> tag on the form, indenting all form contents, hence the strange diff you see here on github.

### DIFF
--- a/printer_zpl2/models/printing_label_zpl2.py
+++ b/printer_zpl2/models/printing_label_zpl2.py
@@ -39,6 +39,9 @@ class PrintingLabelZpl2(models.Model):
         comodel_name='printing.label.zpl2.component', inverse_name='label_id',
         string='Label Components',
         help='Components which will be printed on the label.')
+    menu_ir_values_id = fields.Many2one(
+        comodel_name='ir.values', string='More Menu entry',
+        readonly=True, help='More menu entry.', copy=False)
 
     @api.multi
     def _generate_zpl2_components_data(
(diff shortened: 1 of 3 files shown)


printer_zpl2/models/printing_label_zpl2.py
```diff
@@ -186,3 +189,32 @@ def print_label(self, printer, record, page_count=1, **extra):
...
+    def unlink_action(self):
+        """ Remove the contextual actions created for this label. """
+        self.check_access_rights('write', raise_exception=True)
+        for label in self:
+            if label.menu_ir_values_id:
+                try:
+                    label.menu_ir_values_id.sudo().unlink()
+                except Exception:
+                    raise exceptions.UserError(
+                        _('Deletion of the action record failed.')
```
Garamotte, line 218: Could you add the name of the label when it fails ?
It will help in case we try to remove the menu from several labels at once.

printer_zpl2/models/printing_label_zpl2.py
```diff
@@ -186,3 +189,32 @@ def print_label(self, printer, record, page_count=1, **extra):
...
+            })
+            label.write({'menu_ir_values_id': ir_values.id})
+        return True
+
+    @api.multi
+    def unlink_action(self):
+        """ Remove the contextual actions created for this label. """
+        self.check_access_rights('write', raise_exception=True)
+        for label in self:
+            if label.menu_ir_values_id:
```
Garamotte, line 213: Maybe merge the `for` and `if` statements using:
`for label in self.filtered('menu_ir_values_id'):`

Garamotte (changes requested):
Thanks !
Don't worry about the diff being badly displayed by Github, we all know how to use git for that :)

I tested on runbot, it works, but the menu's name is always "Print Label", without the label name, which seems weird with several labels on the same model: we may want to add a menu item for each label, instead of duplicated items.
I think the best is to duplicate the action instead of simply linking it (and delete it when removing the menu item).This way, you can customize the new action's name (to include the label's name in the displayed menu item) and context (to select this label by default).

Also, a small unit test would be nice.

JosDeGraeve:
Hello @sylvain-garancher,

As i did it now, there is an inconsistency indeed.

I propose these modifications:
* don't link the ir_values record to the label, because it is not specific for that label
* on the label form, move the 'add button', from the top of the sheet, next to the 'model' field.
* If the model already has a linked ir_values record, the button is replaced with a 'remove button'.

That would fix my use case to be more consistent.

What do you think ?

Garamotte:
This is also an option.

I proposed to add label-specific action, you propose to add a label-generic action.
Both are valid, and I have no preference about that, so do it the way you prefer :)
Or maybe someone else has an opinion ?

PR #95, opened 2017-07-11
[REM] Removed dead code

Garamotte:
That's something I wasn't sure during the v10 migration, but these two monkey-patched methods were not called anymore in any module (standard or OCA), and are removed from the next Odoo version.

I found only [one module](https://github.com/OCA/reporting-engine/blob/10.0/base_report_assembler/report_assembler.py#L74-L77) calling these methods, but it hasn't been modified since more than 3 years.

@pedrobaeza @gurneyalex 
Travis is red since this afternoon, but I don't think the issue comes from this repository, did anyone make a change in the Travis or docker configuration ?


Garamotte:
I also updated the pycups version, as this is the version I use everywhere, I thought this could fix the red Travis issue, but no luck...

Garamotte:
I'll rebase on 10.0 when #97 is merged (unit tests fix for mock 2.0.0).

Garamotte:
Any reviews ? :)

cc @pedrobaeza @lasley

pedrobaeza:
OK, seems that the module is not migrated to 10.0, so I suppose there's another way for doing that merging. Please change the version number.

Garamotte:
Done, I changed the last digit.

pedrobaeza:
Why is Travis red?

Garamotte:
Don't know, this seems to be the same issue we already had on other repos, with Travis failing into a timeout sometimes.

Garamotte:
Ok, I reproduced the timeout issue locally. This happens when there is only one database for the test instance (or using db-filter to hide other dbs).

When the tests try to really generate reports, Odoo's HTTP thread is started, but wkhtmltopdf seems to stay blocked waiting for answers, so I think that the HTTP thread doesn't succeeds to start.
When there are several databases (the usual case on my dev box), Odoo's HTTP thread directly returns a 404 error, and wkhtmltopdf isn't blocked.

Did anyone already had that before I spend more time testing on that ?

pedrobaeza:
Thanks for the insights, Sylvain. Maybe @lasley or @moylop260 can say about that, but I haven't experienced that problems.

lasley:
Whoa now that's a new one for me too. Looks like we're maybe hitting wkhtmltopdf/wkhtmltopdf#1888 (or lack thereof) if I am understanding correctly?

Garamotte:
With the help of `kill -3`, I found that the Odoo HTTP workers are blocked at this line:
```
File: "/home/alnsgara/Projects/demo/server/odoo/modules/registry.py", line 55, in __new__
  with cls._lock:
```

I think that this is because the tests are run during the module update, so the registry is not already "fully loaded". The lock is released at the end of the modules update.
Interrestingly, with `--workers=1` (so not in threaded mode, but with only a single HTTP process), the blocking doesn't appear, because the `_lock` attribute is defined as `_lock = threading.RLock()`.
I don't know if it can be considered as a solution for the timeout issue...

Edit: When we have several databases, the "non-db routing" doesn't load any HTTP route, and Odoo automatically returns a 404 for any received request, so no db registry is loaded, and no lock is encountered.

Garamotte:
I finally found how to avoid the timeout with a simple modification in the tests.

There are two methods that set/unset a dummy lock for tests: https://github.com/odoo/odoo/blob/10.0/odoo/modules/registry.py#L428-L444
These methods are used in the `HttpCase` class:
https://github.com/odoo/odoo/blob/10.0/odoo/tests/common.py#L246-L248
https://github.com/odoo/odoo/blob/10.0/odoo/tests/common.py#L263-L265

So I just replaced the use of `TransactionCase` by `HttpCase` in the tests, and everything now works without being block.

pedrobaeza (approved):
:clap:

lasley (approved):
Excellent debugging @sylvain-garancher

PR #96, opened 2017-07-12
[REF] requirements.txt: Use mock==1.0.1

moylop260:
Now [odoo-10.0 use mock==2.0.0](https://github.com/odoo/odoo/blob/68934c6b8/requirements.txt#L13)
But this project is compatible just with mock==1.0.1 

Using the ssh runbot connection to reproduce the travis error the result was:
 - screen shot 2017-07-12 at 13 28 27

Related with https://github.com/OCA/web/pull/671#issuecomment-314850720


Garamotte:
Oh, good catch !
I installed mock 2.0.0 on my local test instance, and now I can reproduce the error.

I'll try to fix the tests, this is better than downgrading a standard requirement, IMO.

Garamotte:
Tests fixed, see #97 
You should be able to cancel this PR.

PR #102, opened 2017-08-09
[PROOF OF CONCEPT] Auto print DESIGN FEEDBACK ONLY - No Merge

gdgellatly:
Hi,

NOTE: This is far from finished or finalized - is for design feedback.  

## Background
In 2010, long before OCA/queue and base_report_to_printer I wrote a module to automatically trigger printing on changes in workflow based on some attributes of a report.  Was pretty basic stuff, used subprocess to manually call lp, relied on CUPS instances for tray selection, a view to handle errors and reprints and a cron to manage queues to speed user response time.  But it worked well, and at v6 and 7 base_report_to_printer didn't really do what it needed to do so I just kept porting.  However, today doing a v10 migration I decided to revisit and just hacked away a few hours to see if it could be done.

## Use case
Base report to printer is user-centric, however when automatically printing documents as part of a 'workflow' it is often a documents properties that matter.  e.g. There is no point printing New Yorks pickings in the LA warehouse even if a dodgers fan loaded the order. 

## Design
So how to mix what I did with base_report_to_printer.
Cool things to know - inactive users can still be used in sudo.

In working, this is how it currently works.  Rather than explain I'll just use screenshots, its clearer.
1. Add automatic user (eventually inactive, but I need to inherit a base_report view first), conditions for use, and print behaviour to a report
image
2. Add a server action
image
3. Add an automated action
image

By default - jobs are set to delay as the view is that the user is usually not interested in an immediate print and indeed may be half a world away from the print destination.

Design Issues:

- First, is this something OCA is interested in, or is it just my unique case?
- I think eventually there will be for example sale_auto_print, mrp_auto_print etc - its just all bundled in there atm for getting POC together.
- Currently it relies mostly on many2many field names for filtering, I'd like for these to be specified dynamically but I am not sure how.  At the very least irrelevant fields should be hidden (again not sure how).  
- I'll probably make company_id a many2one field.
- The fields to filter on are specified in the function call - I did this because at the model level it made no sense (different reports may filter on different fields, or the same report in say different state.  transitions) and I couldn't think of a sane unconfusing way to do that at report level.  Also for this reason I departed from usual field names.
- The setup seems clumsy, is there a better way to achieve?
- One of the issues that led to this approach was base_report_to_printer having no knowledge of the records it is being asked to print, I'm OK with the user-centric part and sudo, but if enhancing base_report_to_printer so it has knowledge of the records coming its way is deemed better, I'm OK with that as well.

There is lots more but I kind of feel this is real long already.

Really anything else.

### DIFF
--- a/base_auto_print/models/auto_print_rule.py
+++ b/base_auto_print/models/auto_print_rule.py
@@ -1,0 +1,71 @@
+from odoo import models, fields, api
+from odoo.addons.queue_job.job import job
+
+
+class Report(models.Model):
+    _inherit = 'report'
+
+    @job
+    @api.model
+    def _print_document_automatically(self, report_name, records, filters):
+        Report = self.env['report']
+        report = Report._get_report_from_name(report_name)
+        PrintRule = self.env['auto.print.rule'].sudo()
+        for record in records:
+            search_args = [('action_id', '=', report.id)]
+            for field in filters:
+                if field in record._fields:
+                    search_args.append(
+                        getattr(PrintRule, '_handle_field_%s' % field,
+                                PrintRule._field_handle_default)(record, field))
+            rule = PrintRule.search(search_args)
+            # or should we fail here, what if user has no rule
+            sudo_uid = rule[0].user_id.id if rule else self.env.uid
+            Report.sudo(sudo_uid).print_document([record.id], report_name)
+        return True
+
+    @api.model
+    def print_document_automatically(self, report_name, records, filters,
+                                     delay=True):
+        if delay:
+            self.with_delay()._print_document_automatically(
+                report_name, records, filters)
+        else:
+            self._print_document_automatically(report_name, records, filters)
+
+
+class IrActionsReportXml(models.Model):
+
+    _inherit = 'ir.actions.report.xml'
+
+    auto_print_rule_ids = fields.One2many(
+        comodel_name='auto.print.rule',
+        inverse_name='action_id',
+        string='Automated Print Rules',
+    )
+
+
+class AutomatedPrintCriteria(models.Model):
+    _name = 'auto.print.rule'
+
+    action_id = fields.Many2one(
+        comodel_name='ir.actions.report.xml', required=True)
+    user_id = fields.Many2one(comodel_name='res.users', string='Print as User')
+    product = fields.Many2many(
+        comodel_name='product.product', string='Products')
+    company = fields.Many2many(comodel_name='res.company', string='Companies')
+    warehouse = fields.Many2many(
+        comodel_name='stock.warehouse', string='Warehouses')
+    product_category = fields.Many2many(
+        comodel_name='product.category', string='Product Categories')
+
+    @api.model
+    def _field_handle_default(self, record, field):
+        self.ensure_one()
+        return field, 'in', record[field].ids()
+
+    @api.model
+    def _field_handle_product_category(self, record, field):
+        self.ensure_one()
+        return ('product_id.product_categ_id', 'in',
+                record.product_id.product_categ_id.ids())
--- a/base_auto_print/__init__.py
+++ b/base_auto_print/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/base_auto_print/__manifest__.py
+++ b/base_auto_print/__manifest__.py
@@ -1,0 +1,14 @@
+# noinspection PyStatementEffect
+{
+    'name': 'Automatic Print Rules',
+    'version': '10.0.1.0.0',
+    'author': 'O4SB',
+    'website': 'http://www.openforsmallbusiness.co.nz',
+    'depends': ['base_report_to_printer', 'queue_job', 'stock', 'mrp', 'sale'],
+    "summary": 'This module allows to select printer based on properties of '
+               'record being printed',
+    'data': [
+             'views/ir_actions_report_xml.xml',
+             'security/ir.model.access.csv'],
+
+}
--- a/base_auto_print/models/__init__.py
+++ b/base_auto_print/models/__init__.py
@@ -1,0 +1,1 @@
+from . import auto_print_rule
--- a/base_auto_print/security/ir.model.access.csv
+++ b/base_auto_print/security/ir.model.access.csv
@@ -1,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_auto_print,Read Auto Print Rules,model_auto_print_rule,base.group_user,1,0,0,0
+manage_auto_print,Manage Auto Print,model_auto_print_rule,base_report_to_printer.printing_group_manager,1,1,1,1
--- a/base_auto_print/views/auto_print_rule.xml
+++ b/base_auto_print/views/auto_print_rule.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0"?>
+<odoo>
+
+  <record model="ir.ui.view" id="auto_form">
+    <field name="name">printing.report.xml.action.form</field>
+    <field name="model">printing.report.xml.action</field>
+    <field name="arch" type="xml">
+      <form string="Report Printing Actions">
+        <group col="2">
+            <field name="user_id"/>
+            <field name="action"/>
+            <field name="printer_id" select="1"/>
+        </group>
+      </form>
+    </field>
+  </record>
+  <record model="ir.ui.view" id="auto_print_rule_view_tree">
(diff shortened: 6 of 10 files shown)


base_auto_print/models/auto_print_rule.py
```diff
@@ -0,0 +1,71 @@
...
+class AutomatedPrintCriteria(models.Model):
+    _name = 'auto.print.rule'
+
+    action_id = fields.Many2one(
+        comodel_name='ir.actions.report.xml', required=True)
+    user_id = fields.Many2one(comodel_name='res.users', string='Print as User')
+    product = fields.Many2many(
+        comodel_name='product.product', string='Products')
+    company = fields.Many2many(comodel_name='res.company', string='Companies')
+    warehouse = fields.Many2many(
```
Garamotte, line 57: Should be named `warehouse_ids` to follow conventions.

base_auto_print/models/auto_print_rule.py
```diff
@@ -0,0 +1,71 @@
...
+
+class AutomatedPrintCriteria(models.Model):
+    _name = 'auto.print.rule'
+
+    action_id = fields.Many2one(
+        comodel_name='ir.actions.report.xml', required=True)
+    user_id = fields.Many2one(comodel_name='res.users', string='Print as User')
+    product = fields.Many2many(
+        comodel_name='product.product', string='Products')
+    company = fields.Many2many(comodel_name='res.company', string='Companies')
```
Garamotte, line 56: Should be named `company_id` to follow conventions.

base_auto_print/models/auto_print_rule.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+
+class AutomatedPrintCriteria(models.Model):
+    _name = 'auto.print.rule'
+
+    action_id = fields.Many2one(
+        comodel_name='ir.actions.report.xml', required=True)
+    user_id = fields.Many2one(comodel_name='res.users', string='Print as User')
+    product = fields.Many2many(
```
Garamotte, line 54: Should be named `product_id` to follow conventions.

base_auto_print/models/auto_print_rule.py
```diff
@@ -0,0 +1,71 @@
...
+        comodel_name='product.product', string='Products')
+    company = fields.Many2many(comodel_name='res.company', string='Companies')
+    warehouse = fields.Many2many(
+        comodel_name='stock.warehouse', string='Warehouses')
+    product_category = fields.Many2many(
```
Garamotte, line 59: Should be named `product_category_ids` to follow conventions.

Garamotte (commented):
Very few comments about conventions in the code (didn"t test).

I think this is a very interresting new feature.
I'm OK to split it in a separate module(s) because of new dependencies.

If the `ir.actions.report.xml` is not sufficient, would it be possible to define a new type of `ir.actions` for automatic printing, instead of defining a python code server action. This new action type could be included in the `base_report_to_printer` module.

gdgellatly:
Hi @sylvain-garancher 

Thanks for looking.  I'm about to restart this next week properly.  The reason I didn't follow convention was twofold.

1. They are many2many fields so ids would be the convention, yet they are primarily many2one fields on the calling object, so I felt it confusing to have ids or id there.  Does automated action call id or ids?

2. The second reason is unlike a normal field, where you have a string label in a form view, these are designed to be passed as a list as part of an automatic action, dynamically matching so it seemed more natural to me to pass as ['warehouse', 'company', ...] in the automated action and less confusing than writing some wrapper code to expand the arguments (less confusing for user anyway, the wrapper code is easy enough to write)

I mean if it was a real blocker, I'd just write the wrapper to append ids and leave the call the same, but this was one case where I felt the convention was more confusing than breaking it.  

But ultimately, I'd like to get rid of these fields entirely, and some how introspect the model so users can configure whatever conditions they want.  Not sure if fields.Reference can do it or not.  I suppose with some funky computed domain that pulls back m2o fields it might work or else maybe an ir.filters object.

Garamotte:
This is not a blocker, your second reason (not confusing the user) is understandeable, but I think using a wrapper could be better (we all think using conventions when reading a field name :D).

About removing these fields, I think adding an "automatic printing" type on `ir.actions.server` could be a good option. This will allow to put the printing action directly on the automated action (without writing python code), and the specific user configuration would be moved on that model, maybe using an `ir.filter`.

Using an `ir.filter` is more customizeable, but maybe sometimes complex for end users.

gdgellatly:
@sylvain-garancher Thanks again,  I've hit a blocker working with master, removal of report module and lots of old report related code, so I'm going to need to do a reasonable refactor of the base printer module.

Doesn't look too bad at this stage, and means I can actually make this module a lot simpler, so I'll do an update next week when I can dedicate more time.

With the way the new domain widget works in v11 I think we will be good to use ir.filters.  If anything it is more natural than the current way.  Indeed, then I can completely rework this to be much better and full featured without massively abusing the users table.

lasley:
@gdgellatly - Interesting indeed. Maybe I'm missing something, but I'm curious why the standard base module isn't working for you in this instance? 

We are currently running a few crons that are able to print automatically without issues. We use a similar strategy to what you laid out, with the crons just running under the service account. As long as the account has the printers set properly, things work well.

The thing that is totally clunky though is routing of specific records or workflows to specific printers. We do this now just with different cron users, but it's definitely not scalable. Is that what you are solving here?

Automatic user - I dunno if I'd recommend using an inactive user for service accounts. Things get totally wonky with the company handling - although I can't remember the specifics. 

What we do for service accounts is to create a user and a group specifically for that user's service privileges. You assign the service account only to the service group, but do not make the service account a share user. Also make sure not to set a password on the user, unless you need API access.

Following this strategy is good in terms of security because the service account only has access to the specific rights that you have predetermined the application needs. Not marking the user as a share user though means that it won't be counted towards an EE/SaaS user limit.

gdgellatly:
@lasley That is pretty much it.

Our previous implementation was very much along your lines.  A little different but general concept.

We had a custom message queue implementation and some models and rules to route to a direct lp command.  Very home made.

So when I went to port it, I decided to try and reuse base_report_to_printer, printer_tray and oca_queue.

The way I handle your main issue (multicompany) is pretty simple.  The inactive users only belong to one company, and really they are just a bridge between some automated printing filters and automatic print users, where we simply say, What properties does the document I'm being asked to print have? Which "service user" is responsible? Print as that user, using the rules defined by base_report_printer.  

But if service accounts do as you describe, then actually there is no practical difference for me between them, except, inactive ones aren't seen by the users ever.  Because of the way base_report_to_printer works, only the user is known at the time the action is selected.  Moreover, on multiple documents it is feasible that each goes to a different printer (i.e requires a different user).

Anyhow for your curiousity our main automated printing workflow goes like this:
When a user confirms an order (jit)

- A sale order (well a document generated by the sale order) is printed at a central quality check location which will vary by shop in v10, warehouse in v11.
- Usually a picking will print at the same location, from a different tray on a different colour (although sometimes it can be a different city entirely). - EDIT: I just remembered this also depends on how its being delivered and the length and weight e.g. Collect, Already taken, Our trucks, intercompany trucks, third party freight, are transport permits required, cranes, booms etc)
- Manufacturing orders dependent on location, product, category could print at a workcentre, or at that same location dependent on whatever, doesn't matter, but usually, it is at that same quality check location on a different colour.

This is across say 9 companies, 12 cities and 70 print queues currently, and I expect the number of printers to double. 

The trick for QC is keeping those documents ordered and together, managing order changes etc, so order becomes important.  One (very busy) QC location could have 50 sales with 200 manufacturing orders an hour and they might not clear the printer for 3 or 4 hours.  You want the documents related to the same sale together if they are just coming out of different trays on the same printer, but for a load of reasons (e.g. putting something on backorder, creating a non sale related transfer or manufacturing) you can't just do it all on the sale, each document needs its own printing logic.  

So we run a queue to make sure things happen in order (currently triggered by a cron).

However, obviously there is some clunkiness to this initial implementation, however, when I look at v11, the entire js part of base_report_to_printer needs to change as the models/methods no longer exist, so there may be an opportunity to modify that call so it sends more context (the records to print I guess) or even if it doesn't modify the function to allow it as a kw argument and then store say an ir.filters directly on the user automation lines, rather than using a bridging user account and model.  But even if it doesn't, the current models are OK.

I think I've determined that ir.filters is the right way to do the filters, maybe with company_id as an exception.  To be honest, the part I am currently stuck on is the js part of base_report_to_printer for v11, so I'd kind of parked this until October when hopefully someone else migrates it.

gdgellatly:
Closing, replacement to come completely reconfigured

PR #103, opened 2017-08-13
base_report_to_printer: Remove temp file after printing

jbaudoux:
The created temp file should be removed once the print job is created.

### DIFF
--- a/base_report_to_printer/printing.py
+++ b/base_report_to_printer/printing.py
@@ -161,6 +161,7 @@ class PrintingPrinter(models.Model):
                              file_name,
                              options=options)
         _logger.info("Printing job: '%s' on %s" % (file_name, CUPS_HOST))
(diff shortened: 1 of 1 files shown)


Garamotte:
Hello,
I agree with you, but the issue is that doing this sometimes deletes the file before it has been printed (I don't remember the exact conditions).

I already have [a commit](https://github.com/syleam/report-print-send/commit/1b906d75ca3ea5b9a70f18679307f05463859b94) to replace the temporary file by CUPS' streaming API, but the current pycups version has a bug with binary data (converted to sort of *ascii*). The fix is already accepted, but no new version has been released, so we cannot install it from pypi. I'm waiting for a new release to propose a PR with this commit.

jbaudoux:
ok, then I'm closing this PR

PR #105, opened 2017-09-06
[IMP] Add QRcode in printer_zpl2 module

sla-subteno-it:
Add new barcode (QR Code format) in printer_zpl2 module


printer_zpl2/models/printing_label_zpl2_component.py
```diff
@@ -142,3 +143,36 @@ class PrintingLabelZpl2Component(models.Model):
     block_left_margin = fields.Integer(
         string='Left Margin',
         help='Left margin for the second and other lines in the block.')
+    qrmodel = fields.Selection(
+        selection=[
+            (zpl2.QRMODEL_ORIGINAL, 'Original'),
+            (zpl2.QRMODEL_ENHANCED, 'Enhanced'),
+        ], required=True, default=zpl2.QRMODEL_ENHANCED,
+        help='Model')
+    qrmagnification = fields.Selection(
```
Garamotte: Replace by a simple Integer field and rename to `magnification_factor`.

printer_zpl2/models/printing_label_zpl2.py
```diff
@@ -84,6 +84,7 @@ def _generate_zpl2_components_data(
         for (component, data, offset_x, offset_y) in to_print:
             component_offset_x = component.origin_x + offset_x
             component_offset_y = component.origin_y + offset_y
+
```
Garamotte: Remove the unneeded empty line.

printer_zpl2/models/printing_label_zpl2.py
```diff
@@ -127,6 +128,23 @@ def _generate_zpl2_components_data(
                     label_data, data,
                     label_offset_x=component_offset_x,
                     label_offset_y=component_offset_y)
+            elif component.component_type == zpl2.BARCODE_QRCODE:
+                # Adding Control Arguments to QRCode data Label
+                data = 'MM,A{}'.format(data)
```
Garamotte: The section (`else`) already calls the `barcode_data` method.
Move the custom data change into the `else` to avoid duplicating code.

You will simply have to add the new arguments added by the QR Code format in the arguments list, as the `barcode_data` already filters the arguments to take in account depending on the barcode type.

printer_zpl2/models/printing_label_zpl2_component.py
```diff
@@ -142,3 +143,36 @@ class PrintingLabelZpl2Component(models.Model):
     block_left_margin = fields.Integer(
         string='Left Margin',
         help='Left margin for the second and other lines in the block.')
+    qrmodel = fields.Selection(
```
Garamotte: Rename to `model`.

printer_zpl2/models/printing_label_zpl2_component.py
```diff
@@ -142,3 +143,36 @@ class PrintingLabelZpl2Component(models.Model):
...
+            (zpl2.QRMAGNIFICATION_400DPI, '400 Dpi'),
+            (zpl2.QRMAGNIFICATION_500DPI, '500 Dpi'),
+            (zpl2.QRMAGNIFICATION_600DPI, '600 Dpi'),
+            (zpl2.QRMAGNIFICATION_700DPI, '700 Dpi'),
+            (zpl2.QRMAGNIFICATION_800DPI, '800 Dpi'),
+            (zpl2.QRMAGNIFICATION_900DPI, '900 Dpi'),
+            (zpl2.QRMAGNIFICATION_1000DPI, '1000 Dpi'),
+        ], required=True, default=zpl2.QRMAGNIFICATION_300DPI,
+        help='Magnification Factor')
+    qrerrorcorrection = fields.Selection(
```
Garamotte: Rename to `error_correction`.

printer_zpl2/models/printing_label_zpl2_component.py
```diff
@@ -142,3 +143,36 @@ class PrintingLabelZpl2Component(models.Model):
...
+        selection=[
+            (zpl2.QRERRORCORRECTION_ULTRA_HIGH, 'Ultra-high reliability level'),
+            (zpl2.QRERRORCORRECTION_HIGH, 'High reliability level'),
+            (zpl2.QRERRORCORRECTION_STANDARD, 'Standard level'),
+            (zpl2.QRERRORCORRECTION_HIGH_DENSITY, 'High density level'),
+        ], required=True, default=zpl2.QRERRORCORRECTION_HIGH,
+        help='Error Correction')
+    qrmask = fields.Integer(
```
Garamotte: Rename to `mask_value`.

printer_zpl2/views/printing_label_zpl2.xml
```diff
@@ -81,6 +81,15 @@
                                     <field name="truncate"/>
                                 </group>
                             </page>
+                            <!-- QRCode arguments -->
+                            <page string="QRCode Arguments" attrs="{'invisible': [('component_type', '!=', 'qrcode')]}">
```
Garamotte: Change the string to `2D Barcode Arguments` to be more generic.

Garamotte (changes requested):

Garamotte:
All changes done.

@pedrobaeza The CLA should be OK, can the bot recheck ?

pedrobaeza:
With a new commit from the person a message will be put about the validity

Garamotte (approved):

aurimukas:
@pedrobaeza could you validate this pull request as well please ?

pedrobaeza:
Sorry, but I don't know about this module. Other should be the reviewer.

Garamotte:
@JosDeGraeve Maybe you can look at this and test the PR ?

JosDeGraeve:
@sylvain-garancher i'll try - please remind me if it takes too long :)

fmdl (approved):

sla-subteno-it:
Thanks @fmdl

aurimukas:
Hello @pedrobaeza can we merge the request please ?

pedrobaeza:
The problem I see with this is that current working installations with zpl2 version 1.0 won't work after this update, which goes against same version stability.

What do you think, @sylvain-garancher ?

Garamotte:
@pedrobaeza The requirements file has been updated, so this is an issue only if the install doesn't respect the requirements, is this still blocking for you ?

pedrobaeza:
Yes. Someone that has a deployment working, updating this repository will find that it's not working anymore.

sla-subteno-it:
@pedrobaeza What solution do you propose to accept this PR?

lmignon:
@pedrobaeza It's a nice improvement, we must find a way to not block this kind of change. IMO it's not blocking but the change of the minimal version required for the module should be mentioned into the README.

lmignon:
@alnslang Can you also modify the setup.py file of this module to enforce the minimal version of spl2 plz as follow:
https://github.com/OCA/report-print-send/blob/10.0/setup/printer_zpl2/setup.py
```python
setuptools.setup(
    setup_requires=['setuptools-odoo'],
    odoo_addon={
        'external_dependencies_override': {
            'python': {
                'zpl2': 'zpl2>=1.1',
            },
        },
    },
)
```
With this change the upgrade will be automatic for those using pip to install the addons.
You should also mention the minimal version required for this external dependency into the README file.

lasley (commented):
Thank you for the improvemenet @alnslang 

In the case of a library version mismatch, can you explain to me what the error is? From reading the code, I am guessing that the error would trigger any time that a label is printed? Am I correct in that guess?

pedrobaeza:
OK, then go ahead. I'm not a direct user of this module, but I was trying to avoid breaking installations. If you say that there's no problem or it's an acceptable risk, then I won't block anymore.

printer_zpl2/README.rst
```diff
@@ -15,6 +15,7 @@ See below for more details.
 Installation
 ============
 
+This module requires zpl2 >= version 1.1.
```
lasley: Instead of just noting the requirement, we need to state this as an upgrade note. Users upgrading pre-existing installations are the ones that this affects, not new users. As such, I recommend:

```
**Upgrade Note:** If upgrading from a version prior to 10.0.2.0.0, you must make sure to update the `zpl2` Python package to at least version 1.1.
```

Also, please bump the module version by a major version: 10.0.2.0.0

PR #109, opened 2017-10-03
[MIG][11.0] Migrate base_report_to_printer

Garamotte:
Migration of the `base_report_to_printer` module to v11.0 (currently tested on master as the 11.0 branch has not already been created on Odoo's repo).


gdgellatly:
@sylvain-garancher A couple of lint errors, like forgetting to bump the version number I'll bring it down and test today.

gdgellatly (commented):
@sylvain-garancher 
Full Functional Test on local 11.0 Official and Enterprise after technical review (which was fine)

Minor Points
- Move menu from after General Settings to Before - personally I think it should be after translations
- After adding localhost server, restart of server was required in order to update printers. - Fix or add to known issues
- Once a printer is set default - there is no way to unset so there is no default 
- The set default button should be invisible if the printer is default
- Remove the name label from Printers form - it looks wrong/inconsistent with usual UI
- Cancel all running jobs button should have confirmation dialog IMO
- I think we should be opinionated here and provide a non-debug reports menu entry in the printing menu
- on the Report form view (or associated field string) I think Action and Printer should be renamed Default Behaviour and Default Printer
- The user tree view should be editable
- User and Printer many2ones on this view should have no create or edit.
TBC - I need screenshots for next part

gdgellatly:
@sylvain-garancher 

Traceback with following config printing Purchase Order as Administrator
Error:
Odoo Server Error

Traceback (most recent call last):
  File "/home/graemeg/PycharmProjects/odoo10-c/odoo/addons/web/controllers/main.py", line 1639, in report_download
    response = self.report_routes(reportname, docids=docids, converter='pdf')
  File "/home/graemeg/PycharmProjects/odoo10-c/odoo/odoo/http.py", line 514, in response_wrap
    response = f(*args, **kw)
  File "/home/graemeg/PycharmProjects/odoo10-c/odoo/addons/web/controllers/main.py", line 1588, in report_routes
    pdf = report.with_context(context).render_qweb_pdf(docids, data=data)[0]
  File "/home/graemeg/PycharmProjects/odoo10-c/oca_report_send/base_report_to_printer/models/ir_actions_report.py", line 135, in render_qweb_pdf
    printer.print_document(self, document, self.report_type)
  File "/home/graemeg/PycharmProjects/odoo10-c/oca_report_send/base_report_to_printer/models/printing_printer.py", line 94, in print_document
    os.write(fd, content)
TypeError: a bytes-like object is required, not 'tuple'

image

As a side point - If we rename printer in that screenshot to Default Printer and add help text that specifies that this printer will be used if no printer specified on users I think would be nice.

EDIT: Just a note, I know the directory says odoo-10c - it isn't it's latest checked out Odoo11 Community and Ent sources.

base_report_to_printer/models/printing_printer.py
```diff
@@ -0,0 +1,174 @@
...
+    def print_document(self, report, content, format, copies=1):
+        """ Print a file
+
+        Format could be pdf, qweb-pdf, raw, ...
+
+        """
+        self.ensure_one()
+        fd, file_name = mkstemp()
+        try:
+            os.write(fd, content)
```
gdgellatly, line 94: Changing this to content[0] resolved traceback, and printed document, however it also printed on screen as well.

Garamotte: Thanks.
This is because the `render_qweb_pdf` method returns a tuple containing the document (contents) and the format ('pdf' here).
I'll fix it in my next commit ;)

Garamotte:
@gdgellatly Thanks a lot for the review, some points were things I forgot before pushing (like the lint) :)

I kept the review fixes as a separate commit to ease reviewing the changes:
- The Printing menu has been moved before General Settings, but I can't ensure it will always be after Translations: Translations has sequence=1 and General Settings has sequence=2... and I prefer not to change sequences of standard menus here.
- I succeeded to get printers without restarting the Odoo server, right after adding the printing server, even without clicking on "Save", but directly on "Update printers". Can you have more details about the issue on your side ?
- I added an Unset Default button, displayed only when the printer is the default. The Set Default button is now hidden in this case. I didn't succeed to put the two buttons at the same position...
- The label of the name field has been removed on the printers form view.
- Adding a confirmation dialog on the cancel all jobs button is a very good idea. I think this change should be backported to v10.0 ;)
- The non-debug mode "Report" menu wasn't here because I commented it, and forgot to search for the new name of the action, this is fixed now.
- "Action" and "Printer" have been renamed to "Default Behaviour" and "Default Printer" on the report form.
- I don't understand the user tree view.

I didn't added the no create on many2one fields yet, but added it on my TODO list.

gdgellatly:
@sylvain-garancher Sorry for being unclear, I meant the one2many user list view in the reports

- The menu position is fine.
- Its late, I probably did something wrong, maybe it happens if pycups isn't installed when the server is started or something as it was a new Py3.5 venv.
- I wasn't especially worried about the default thing, just writing as I reviewed (although a smart button using widget boolean_button would be the ideal solution here)
- Confirmed field label removed
- Thanks - confirmation dialog confirmed
- Those field names haven't changed, maybe you forgot to commit a file.

Re the poorly explained user tree view, I mean the one2many list on the report view
image

Currently opens to a form
image

I'm still getting both the report on screen and printed, but its late, so if you fixed it, then I'll start afresh tomorrow.

Garamotte:
Oops, you're right, I forgot to commit some changes.
I'll look the on screen print issue later, this is not fixed yet.

gdgellatly:
One last piece on usability.

What is the purpose of Printing User and Printing Manager security groups.  To test I created 2 users.
Test (not a member of either) and Test2 (a member of user)

- Both Test and Test2 could be assigned actions. (which is fine, what I was really testing was that Test2 couldn't change Test defaults)
- Neither Test or Test2 could change their printing settings (EDIT: under User Preferences)
- Neither test or test 2 could do anything I could see related to printing

In fact I could see no difference at all between the access with printing user set and not set.

Garamotte:
I fixed the latest views details you mentionned.
About the access rights, I think this is mainly to define a role (manager) who is able to configure the servers/printers, and the other (user) to be able to use them (print documents).

gdgellatly:
OK, in that case I think all we need to do is add the printing fields to 

```python
SELF_WRITEABLE_FIELDS = ['signature', 'action_id', 'company_id', 'email', 'name', 'image', 'image_medium', 'image_small', 'lang', 'tz']
```
of res.users model like so (copied from mailing module)
```python
def __init__(self, pool, cr):
        """ Override of __init__ to add access rights on printing. 
            Access rights are disabled by default, but allowed
            on some specific fields defined in self.SELF_{READ/WRITE}ABLE_FIELDS.
        """
        init_res = super(Users, self).__init__(pool, cr)
        # duplicate list to avoid modifying the original reference
        type(self).SELF_WRITEABLE_FIELDS = list(self.SELF_WRITEABLE_FIELDS)
        type(self).SELF_WRITEABLE_FIELDS.extend(['printing_action', 'printing_printer_id'])
        # duplicate list to avoid modifying the original reference
        type(self).SELF_READABLE_FIELDS = list(self.SELF_READABLE_FIELDS)
        type(self).SELF_READABLE_FIELDS.extend((['printing_action', 'printing_printer_id'])
        return init_res
```

Garamotte:
I added the fields in the WRITEABLE and READABLE fields lists, and also had to put them explicitely `readonly="0"` on the user's perfs view.

gdgellatly:
@sylvain-garancher I've traced down that bug with both screen and printer report.  In order to replicate you need the setup as in screenshot above.
Default Printer - set
Default Behaviour - send to client

Then a user action of Send to Printer without Printer (or Trays set)

Everything else I tested works fine.

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -0,0 +1,138 @@
...
+            }
+        return result
+
+    @api.multi
+    def print_document(self, record_ids, data=None):
+        """ Print a document, do not return the document file """
+        document = self.with_context(
+            must_skip_send_to_printer=True).render_qweb_pdf(
+                record_ids, data=data)
+        behaviour = self.behaviour()[self]
```
gdgellatly, line 100: @sylvain-garancher I think I've noticed what may be a bug that is never triggered.  Note here that we call behaviour, and then use the self recordset as the index.  Now I can't think of a case where it isn't a singleton, but notice in `def behaviour` we iterate over self, so in theory there is a chance self is not in the result.  Indeed all 3 requests to that function originate from `print_document`, I just wonder if we are better to remove the loop in behaviour, use ensure_one() return a dict and remove the indexing in the 3 calls.  Would be a bit cleaner.

Garamotte: Oops, this is an error I added during the migration, when I merged the `report` and `ir.actions.report.xml` models methods.
The original method was in the `report` model, and searched for a report, then used the report's id: https://github.com/OCA/report-print-send/blob/10.0/base_report_to_printer/models/report.py#L11-L23

I fully agree to add `self.ensure_one()` in this method. This is indeed always a singleton, because it was a single `report_name` argument before.

gdgellatly: @sylvain-garancher you might want to check my refactor at #112  which does this and completely reimplements some of the quirks around printer options which blocked me doing #111.  Its all rebased off this PR plus #110 .  It's just I'm not sure where to go, you've done all this port work, I've done a load of work on top of your port.  I kind of feel like the PR's should be reviewed on their own merits, but then merged together.

I have one more bug/design issue to consider.  If you create a user on 11.0 enterprise and do not give them the printing user security group you get a traceback on logging in.  

However, I wonder if there is really any point to this security group anyway and we don't just use base.group_user (or whatever it is now), or at least make it an implied group of that group.  

If I think about it, we can't stop a user printing a report (to screen) and clicking ctrl-p to print or emailing it.  In terms of accessing a report, well reports already have security.  It just feels like the group isn't really doing anything of value, and is very easy to miss when creating users.  

I struggle to think of a use case where a user is given a device to access an ERP, but not to use a printer.  If anything the security detail might be at the printer or server level.

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -0,0 +1,138 @@
...
+            result[report] = {
+                'action': action,
+                'printer': printer,
+            }
+        return result
+
+    @api.multi
+    def print_document(self, record_ids, data=None):
+        """ Print a document, do not return the document file """
+        document = self.with_context(
```
gdgellatly, line 97: @sylvain-garancher I'm becoming a bit anti these kind of calls for tightly coupled behaviours.  It makes inheritance difficult as you can't pass a context down a chain of calls.  In this example we could be passing anything from languages, to keys used in the report.  In any case I see no reason not to send the passed context as well.  I'd rather we had something like

```
ctx = dict(self._env.context, must_skip_send_to_printer=True)
document = self.with_context(**ctx)...
```

Obviously not quite like that, just the idea, send the original context as well.

Garamotte: I'm not sure to understand your concern, but I hope this answers:
```
>>> self.env.context
{'tz': False, 'lang': 'en_US'}
>>> self.with_context(must_skip_send_to_printer=True).env.context
{'tz': False, 'lang': 'en_US', 'must_skip_send_to_printer': True}
```

gdgellatly: Hi

@sylvain-garancher try again but this time with a different context other than default. 

Your example is not adding the existing context, it is merely recreating a default context in the new one.  So any context passed is lost when sending to printer because we are creating a new environment.

So 
```python
self.with_context(example='example').print_document(...)
example won't be passed to the new context.
```

gdgellatly: @sylvain-garancher weird thing is I tested this for ages and it didn't work.  I just tested now in shell and it did work.  I'll test again properly later, ignore for now.

Garamotte: Maybe you tested using a dictionnary as argument ?
When we use a dictionnary as positional argument (`self.with_context(new_context)`), the context is fully replaced by this dict.
But if we give keyword arguments, each argument is added in the current context (replacing the already existing key, if any).

gdgellatly: @sylvain-garancher you are probably right.  I can't fully remember now, that was late nights back then, but it was adding arguments that were returned i a dictionary format although I thought I'd **ed the return, but I just tested it and it works EDIT: To clarify, passing a **dict() adds to context not replaces.

In truth its great for me I am wrong.  Just allowed me to remove a ton of useless override.

gdgellatly:
@sylvain-garancher et al, lets work to get this done.  If you are busy I can take over.  I think its nearly there.

Garamotte:
Yes, sorry, I'm very busy at the moment...
You can fix the latest things discussed here in your PR that merges `printer_tray` into `base_report_to_printer`, and I'll close this one.

gdgellatly:
@sylvain-garancher I missed this, anyway I was just doing something else and I noticed they refactored the web client post release, so the js is going to need to change now anyway as there is no more web.Model

gdgellatly:
@lasley Can I ask you for a quick bit of help please.  In this module we had a javascript call to select the action, either print or send to client.  With changes to the web client it is no longer valid and I can't work out how it needs to be changed to still work.  Do you know how to call a models method in 11.0 web client?

lasley:
@gdgellatly - Sure, can you point me to the old code we're trying to replace?

base_report_to_printer/static/src/js/qweb_action_manager.js
```diff
@@ -0,0 +1,44 @@
...
+    ActionManager.include({
+        ir_actions_report: function(action, options) {
+            action_val = _.clone(action);
+            var _t = core._t;
+            var self = this;
+            var _super = this._super;
+
+            if ('report_type' in action_val && action_val.report_type === 'qweb-pdf') {
+                framework.blockUI();
+                new Model('ir.actions.report').
```
lasley, line 18: @gdgellatly Is this what we're replacing?

gdgellatly:
@lasley Yes,

web.Model no longer exists, I tried a straight swap with Basic and Abstract but no dice.  The call method fails so I'm guessing its a bit more complicated than that.  When I debug it, I see that the args are being passed as the main arguments, instead of the method and method args.

lasley:
@gdgellatly - sweet thanks for the elaboration. And when you say you used the basic and abstract, you're referring to `web.BasicModel`?

From what I understand, the models just allow for the basic CRUD operations. You've honestly got me if there is a new entrypoint for any method call, but I think an easy solution here would just be to use RPC directly instead:

```
var rpc = require('web.rpc');
...
rpc.query({
    model: 'ir.actions.report',
    method: 'print_action_for_report_name',
    args: [action_val.report_name]
}).then(...
```

gdgellatly:
@lasley Thanks. Yes its quite different now and completely undocumented and I never really understood the last one, so not having fun.  I'm going with what you did.  That was very close to my hackish attempts.

lasley:
If I understand correctly, things would be different if we actually had a record to attach to. I haven't actually tried it, but I feel like you can `require('web.BasicModel').get(id).a_method(arg1, arg2)`

gdgellatly:
@lasley thanks it worked, I've done up the outstanding details of this PR and done a full review and pushed to #110 as a single PR

Garamotte:
I close this PR to avoid splitting the conversations :)
Superseded by #110

PR #88, opened 2017-05-22
Printer copies

jeanpaulrobineau:
Hello,

This module will allow to specify per report or user/report the number of copies.
This is inspired by printer_tray module

kind regards,
Jean-Paul Robineau

PR #110, opened 2017-10-04
[11.0][MIG] base_report_to_printer + printer tray: Merged together

gdgellatly:
Relies on #109 

Straight port with addition of _order attribute on printer tray
Tested locally


gdgellatly:
Not sure I did the history migration right, if not point me to a doc and I'll redo

Garamotte:
I didn't test localy, but the code review seems OK.
Al commits are here, I think you did things the right way. You should simply rebase to squash oca trabsbot commits.

The reindent of XML files is a good thing (even if it produces a bigger diff).
You may just rename the `tests/test_ir_actions_report_xml.py` to `tests/test_ir_actions_report.py` to follow other renamings ;)

gdgellatly:
OK, rebased, and renamed test. also I added group names on the printing_printer view.  You should consider for like view in #109 to make the inheritance a little cleaner.

gdgellatly:
@sylvain-garancher et al, I think before merge we should have a discussion re merging this module with base_report_to_printer.  The logic is simply that users are used to a screen (e.g. when pressing ctrl-p) that allows them to select printer trays.  They are a concept already central to printing.  

To me that makes the idea of selecting a printer tray being an additional feature that requires installing not make as much sense as including trays in the base module.  

Of course there are some minimal technical benefits (less loc, no glue modules when an extension wishes to use both) and technical costs (forward/backports) but I don't see them as that relevant.

Garamotte:
I agree, this small module could be merged into the base one.
@pedrobaeza confirmed he agrees too.

gdgellatly:
@sylvain-garancher I've given it a go, was a bit of merge and rebase hell with your PR.  I just need to know what to do with translations.

Garamotte:
It seems you failed the `_prepare_update_from_cups` method in `base_report_to_printer/models/printing_printer.py` during the merge: This method now appears twice in the file.

For translations, I don't know if many languages are filled or not.
You may concatenate the files of the two modules (removing the header of the second module for each file), it should prevent losing data, and may be "repared" on the next transbot pass (I didn't check that).

yvaucher:
Indeed, seems reasonable to include this in base module. I basically did it separately because I though it was a specific use case to add that extra option.

pedrobaeza:
About translations, ideal is to merge them, changing module on each line from the other module, but it's a lot of work, so maybe we can just leave the second translation out.

gdgellatly:
FYI - I went to merge them, was nearly finished and then noticed it probably won't work, just because of the number of duplicate strings and how the file is structured, so I think I'll leave them out for now.

gdgellatly:
@sylvain-garancher I forgot to commit 4 files.  Regards the translations - actually just deleted the directory, but I can do the merge of the files if we want to.  Its 4am for me, I'll check the test failures in the morning, at first glance it looks more related to the tests than the code.

base_report_to_printer/models/printing_printer.py
```diff
@@ -0,0 +1,244 @@
...
+    def print_options(self, report=None, format=None, copies=1):
+        """ Hook to set print options """
+        options = {}
+        if format == 'raw':
+            options['raw'] = 'True'
+        if copies > 1:
+            options['copies'] = str(copies)
+        if report is not None:
+            printing_act_obj = self.env['printing.report.xml.action']
+            if report.printer_tray_id:
```
yvaucher: Here you get a string for report instead of a record set, according to tests
https://travis-ci.org/OCA/report-print-send/jobs/283709521#L1157

yvaucher (changes requested):

base_report_to_printer/models/printing_printer.py
gdgellatly: @yvaucher manual debug shows report is a recordset singleton of ir.actions.report.  Can you give me a little guidance fixing the test.  For now I've set it to None in the test, which is a valid value for the keyword argument as the tray functionality tests I kept in a seperate test file to cover that code.

The final error is the KeyError on printer-uri-supported - I'm not sure whether I should be modifying the test vals here, or modifying the code to allow for the missing key.  In manual testing, and prior to merge all tests passed fine, so it gives me the sense that in the real world cups always returns this key, so maybe the test vals?

Garamotte: The test was wrong here, the `'pdf'` string is the `format` argument (forgot to be named), not `report`.

We didn't see the error before because the test is in the base module, and the `print_options` method in that module doesn't use the `report` argument.

Garamotte:
The remaining errors `TypeError: unicode or bytes object required` are mocking issues.
To fix these errors, you'll have to assign a return value to the `getPPD3` method, like this:
`cups.Connection().getPPD3.return_value = (0, 0, '')`.

gdgellatly:
@sylvain-garancher @yvaucher tests all now fixed.  One bug from original base_report_to_printer to fix, but will do as seperate commit.

gdgellatly:
HELP NEEDED: 

There is an issue in the `def print_options` assigning the printer tray value.  It relies on a report_action, yet there is technically no reason that print_file is required to be called by it, and in fact `def behaviour` of `ir.actions.report` is called (twice in fact) directly on the report_action returning the tray already so it is superfluous to do all that searching again.

I think we should either for the specific case just add a new kw argument, or for the more general case (such as PIN protecting printing, print in grayscale or any other valid option) consider a **kwargs with an option dispatcher e.g.

```
def print_options(self, report=None, **kwargs):
    option = {}
    for option, value in kwargs.items():
        func = getattr(self, '_get_option_%s')
        if func:
            options.update(func(value, report=report))
        else:
            options[option] = value
    return options
```

Thats the general idea, whether we deprecate print_options or not idk as we'd need the **kwargs argument on print_document and print_file as well.

gdgellatly:
Hi, I know this PR is old #88 , but is exactly why I proposed #112 .  If you look at the code, there are 2 flaws.

1. behaviour sets copies, first in `print_document` then in '_render_qweb_pdf` then it is never used.
2. `def print_options` never uses behaviour, and creates an implied constraint that print options may only be set by actions.

That is basically what my proposal is to fix.  

The job of the printer class is to send documents to the printer.
As part of that it will pass options, but it is not responsible for providing them (although it may sanitize and parse them)

In that way, we completely decouple printer from actions and reports.

Then it is the job of the various behaviour functions (in report, or report.action OR ANY OTHER model, function etc that wishes to take advantage of the printers resource) to determine what those options are. As an example of such a utility there is this POC, which is fully testable https://github.com/gdgellatly/report-print-send/tree/11.0-refactor-auto-print based on this PR.

gdgellatly:
Just hold on reviewing this.  I think the new way they have implemented the Report Controller means that this javascript code intercepts the wrong route so we end up with both an old style report action and a new style controller resulting in the report printing both on printer and on screen

base_report_to_printer/data/printing_data.xml
```diff
@@ -30,8 +30,10 @@
         <field name="numbercall">-1</field>
         <field eval="False" name="doall"/>
         <field name="model_id" ref="base_report_to_printer.model_printing_server"/>
-        <field eval="'action_update_jobs'" name="function"/>
-        <field eval="'()'" name="args"/>
+        <field name="state">code</field>
+        <field name="code">
+model.action_update_jobs()
```
Garamotte: Thanks, I discovered the change of crons recently, and didn't adapt the `base_report_to_printer` PR.
You can put this on a single line, like I did here: https://github.com/OCA/stock-logistics-barcode/pull/110/commits/b4cc1933958704ee0338c6879cb0677647911244#diff-2cad5a619c492c769aa7bf1f7cb7e9e7R15

gdgellatly: Thanks done

gdgellatly:
@pedrobaeza @lasley @sylvain-garancher I need one last bit of help to get this done (or we call it a feature and put in known issues).

In qwebactionmanager.js of web there is a function called trigger_download which calls the report_download route in web.ReportController.  It is this function that ultimately prints the report and generates the client response.  So what is happening as best I can tell is that our javascript is called and correctly doesn't call super.  But there is something else in the web client that does regardless.

I'm kind of out of ideas on how to approach and fix this.  I've tried a few things within the existing paradigm, such as overiding the report_download function to return nothing (endless loop), changing the action_val report type (no effect, not passed).

Also note there is a further complication, that is that nearly every module in reporting-engine repo overrides this same function and does call super.

I kind of get the feeling that the approach needs a complete rethink or else I'm just missing something obvious.

pedrobaeza:
JS is not my best skill. @Yajo can you help with this?

lasley:
@gdgellatly - I'm having trouble finding the method you're referring to in this code. Can you please comment inline on the method in question?

gdgellatly:
@lasley sure

https://github.com/odoo/odoo/blob/11.0/addons/web/static/src/js/report/qwebactionmanager.js#L110
for the trigger download code

https://github.com/odoo/odoo/blob/11.0/addons/web/controllers/main.py#L1637 for the ultimate route.

EDIT: These are what are called, no matter what the current js does.

base_report_to_printer/README.rst
```diff
@@ -0,0 +1,127 @@
...
+Installation
+============
+
+* Install PyCups - https://pypi.python.org/pypi/pycups
+
+.. code-block:: bash
+
+   sudo apt-get install cups
+   sudo apt-get install libcups2-dev
+   sudo apt-get install python-dev OR sudo apt-get install python3-dev
```
yajo: This should be only `python3-dev`. Python 2 is not supported in odoo v11.

base_report_to_printer/README.rst
```diff
@@ -0,0 +1,127 @@
...
+   sudo easy_install pycups OR sudo pip install pycups
```
yajo: easy_install, really? Drop that please 😆

base_report_to_printer/README.rst
```diff
@@ -0,0 +1,127 @@
...
+The notification when a report is sent to a printer will not be
+displayed for the deprecated report types (RML, Webkit, ...).
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/144/11.0
+
+
+Known issues / Roadmap
+======================
```
yajo: Delete this empty section.

base_report_to_printer/migrations/9.0.2.0.0/post-10-create_server_record.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2016 SYLEAM (<http://www.syleam.fr>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import SUPERUSER_ID, api
+from openerp.tools.config import config
+
+__name__ = 'Create a printing.server record from previous configuration'
+
+
+def migrate(cr, v):
```
yajo: Delete this file, it's a migration for v9

base_report_to_printer/static/src/js/qweb_action_manager.js
```diff
@@ -0,0 +1,49 @@
...
+                        framework.unblockUI();
+                        rpc.query({
+                            model: 'ir.actions.report',
+                            method: 'print_document',
+                            args: [action_val.context.active_ids, action_val.report_name],
+                            kwargs: {data: action_val.data || {}},
+                            context: action_val.context || {}
+                        }).then(function () {
+                            self.do_notify(_t('Report'),
+                                _t('Document sent to the printer ') + print_action.printer_name);
```
yajo: Don't concatenate strings, inerpolate with `%s` and `_.str.sprintf()`, for better translations.

base_report_to_printer/static/src/js/qweb_action_manager.js
```diff
@@ -0,0 +1,49 @@
...
+                        }).fail(function () {
+                            self.do_notify(_t('Report'),
+                                _t('Error when sending the document to the printer ') + print_action.printer_name);
```
yajo: Use `%s` also.

base_report_to_printer/static/src/js/qweb_action_manager.js
```diff
@@ -0,0 +1,49 @@
...
+
+
+    ActionManager.include({
+        ir_actions_report: function (action, options) {
+            var action_val = _.clone(action);
+            var _t = core._t;
+            var self = this;
+            var _super = this._super;
+
+            if ('report_type' in action_val && action_val.report_type === 'qweb-pdf') {
```
yajo: If I'm not wrong, you should be adding a new `action.report_type` called i.e. `qweb-pdf-server`, or make the action have notion of being printed in the server somehow.

Upstream code uses that key to know [if you need an HTML report](https://github.com/odoo/odoo/blob/b37cc1f9b7897759a86cdccfa1b06d58361a97da/addons/web/static/src/js/report/qwebactionmanager.js#L77) or [a PDF report](https://github.com/odoo/odoo/blob/b37cc1f9b7897759a86cdccfa1b06d58361a97da/addons/web/static/src/js/report/qwebactionmanager.js#L88), so you should just do in this method:

1. Check if `action.report_type === "qweb-pdf-server"`.
1. If it's true, do my things.
1. If it's false, `return this._super.apply(this, arguments)` and forget.

yajo (changes requested):
I hope it helps 😊

base_report_to_printer/static/src/js/qweb_action_manager.js
gdgellatly: thanks @Yajo it will take me a while to get my head around.  I don't think a new report type is correct, because every qweb-pdf report is subject to a specific or generic printing behaviour, even if that behaviour is just to send to client.

The current js works, and prints as expected, its just that it also downloads to the client, so super is being called regardless of this function.  Is there some way we can, say around line 33, once we know it is to be printed, change `action_val.report_type` so it falls through?

lasley: Total shot in the dark, but I wonder if [Action.ir_actions_report](https://github.com/odoo/odoo/blob/6d7d6d9460cae90b257daecf848de1873c413448/addons/web/static/src/js/chrome/action_manager.js#L905) is the culprit. This is the first I've looked at this `js/chrome` folder though - I have no idea wtf it's for

gdgellatly: @lasley @Yajo OK I made massive progress.  I found a bug in the js code where action_val was checked instead of action.  As a result, none of this javascript was ever called anyway, yet it still printed the report.  

I just commented out the entire file and still got my printed report (plus the client download).

I then just manually edited the report/download http in web/controllers/main to return this (warning ugly ahead)
```python
    @http.route(['/report/download'], type='http', auth="user")
    def report_download(self, data, token):
        """This function is used by 'qwebactionmanager.js' in order to trigger the download of
        a pdf/controller report.

        :param data: a javascript array JSON.stringified containg report internal url ([0]) and
        type [1]
        :returns: Response with a filetoken cookie and an attachment header
        """
        requestcontent = json.loads(data)
        url, type = requestcontent[0], requestcontent[1]
        try:
            if type == 'qweb-pdf':
                reportname = url.split('/report/pdf/')[1].split('?')[0]
                # <-----------------snip-------------->
                error = {
                'code': 200,
                'message': "Report Sent to Printer",
                    'data': {}
                }
                return request.make_response(html_escape(json.dumps(error)))
```
and I got this (EDIT: And the printed report)

image

so what I think the best strategy is
1. Remove the javascript
2. Override report_download with this algorithm
 - Call super
 - If it is to be sent to printer -> Return notification report has printed else return super response.

So (and sorry for using you guys as support here) how do I return a notification rather than an error in a python web controller?

lasley: @gdgellatly - A notification as in one of the javascript boxes that pops up in the top right corner from time to time? Or some sort of error that still allows processing?

base_report_to_printer/static/src/js/qweb_action_manager.js
```diff
@@ -0,0 +1,49 @@
...
+                    if (print_action && print_action.action_val === 'server') {
+                        framework.unblockUI();
+                        rpc.query({
+                            model: 'ir.actions.report',
+                            method: 'print_document',
+                            args: [action_val.context.active_ids, action_val.report_name],
+                            kwargs: {data: action_val.data || {}},
+                            context: action_val.context || {}
+                        }).then(function () {
+                            self.do_notify(_t('Report'),
```
gdgellatly: @lasley  Essentially I just want to replicate this block in python - so the top right kind of notification

lasley: Damn I thought that was the case - I don't know of a way to do this. The [crash manager](https://github.com/odoo/odoo/blob/cd48a806ac636806130efcf5c055f73758cb0902/addons/web/static/src/js/services/crash_manager.js) is what handles anything we can send back error-wise from Python, and it does dialogs.

lasley: An interesting module idea would be a `NotificationError` in Python and something in JS that intercepts the internal RPC calls, catches errors, and handles appropriately if a notification error. I don't think that exists though.

gdgellatly:
@lasley @Yajo et al, I got it, don't waste any more time helping me.

gdgellatly:
@lasley @pedrobaeza @sylvain-garancher @Yajo Ok Ready for review.  My apologies, I went down a rabbithole with this javascript, which I don't really get, but finally realized the error was in the change to using rpc.

I left it as a seperate commit so you can see how dumb I am.

ougc27 (approved):
Hello

I just tried locally this module successfully :+1:

base_report_to_printer/tests/test_ir_actions_report.py
```diff
@@ -73,6 +73,7 @@ def test_print_action_for_report_name_returns_if_report(self):
         """ It should return correct serializable result for behaviour """
         with mock.patch.object(self.Model, '_get_report_from_name') as mk:
             res = self.Model.print_action_for_report_name('test')
+            del res['id']  # HELP!
```
yajo: What if you add below to `expect` dict: `'id': behaviour.id` instead?

yajo (changes requested):

base_report_to_printer/tests/test_ir_actions_report.py
gdgellatly: I tried it, but I don't think it will work, behaviour is just a method, not an instance of a report.

gdgellatly: actually, now that I think about it, probably the way I did the test is best for now.  print_action_for_report_name adds id and is the only function that adds id becuase it is only needed for the rpc query.  By deleting that key, we are confirming that it is adding that key, and only that key.

gdgellatly: @Yajo I fixed it so its much better

yajo (approved):

pedrobaeza:
@yvaucher please give your final blessing

yvaucher (approved):

yvaucher:
Thanks @gdgellatly great work let's merge it

PR #112, opened 2017-10-08
11.0 refactor options

gdgellatly:
Hi,

In #109 and #110 I made some comments regarding the way we do print options and the faact behaviour is not a very correct function and should just be called self.ensure_one() unless there is some reason it was done to return a dict of dicts.

The guts of this refactor is that I'd like base_report_to_printer to become a bit more like a lot of other base_ modules, where they leave very easy extensibility.  With this change we seperate responsibilities.

`ir.actions.report.behaviour` is responsible for defining print_options, although overrides could occur anyway in the `ir.actions.report` call chain, `printing.printer` is solely responsible for formatting the passed options to a cups compatible dict.

This isn't really for merge, as I've made a mess with working on 5 different branches, but mainly for comment and hoping we can do into the initial release.  Its easy enough to retype from scratch if needed.  It is the last 2 commits (1 code, 1 adjusting tests for changed return signature of behaviour).

Just on the change of behaviour, maybe someone can explain why it returned dict of dicts, was it just for the javascript, although that doesn't make sense as it just peeled self anyway?

EDIT: Sorry one final edit, I should mention the purpose behind this is to greatly extend this module.  Of course there is the automated action workflow printing that is coming soon, but also just to be able to do any future extension we want easily, e.g. pass any option we want, GrayScale, Output Tray, PIN protect, Billing code, non standard PPDs etc.  I have a lot of printers, ranging from 1 month to 25 years old running a range of applications, all automatically printing out of Odoo, unfortunately not all PPD's present options the same way and of course they don't even have all the options sometimes.


gdgellatly:
FYI - the test failures on printer tray are both expected and a little more complex to fix.  They need splitting, so that the behaviour portion of the option is in `IrActionsReport` and the correct formatting is on `PrintingPrinter`, but I'll wait for more general feedback.

There's also a question here of behaviour in general, which requires a little code adjustment to exactly match (i.e. I find it strange to set a tray on the report and that tray is used by a users action, but I'll make it behave the same as old code).

gdgellatly:
Hi, I know this PR is old #88 , but is exactly why I proposed #112 . If you look at the code, there are 2 flaws.

    behaviour sets copies, first in print_document then in '_render_qweb_pdf` then it is never used.
    def print_options never uses behaviour, and creates an implied constraint that print options may only be set by actions.

That is basically what my proposal is to fix.

The job of the printer class is to send documents to the printer.
As part of that it will pass options, but it is not responsible for providing them (although it may sanitize and parse them)

In that way, we completely decouple printer from actions and reports.

Then it is the job of the various behaviour functions (in report, or report.action OR ANY OTHER model, function etc that wishes to take advantage of the printers resource) to determine what those options are. As an example of such a utility there is this POC, which is fully testable https://github.com/gdgellatly/report-print-send/tree/11.0-refactor-auto-print based on this PR.

gdgellatly:
I've merged this into #110

PR #114, opened 2017-10-19
[10.0] Pingen migration

abaghirli:


pingen/__init__.py
```diff
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 ##############################################################################
 #
-#    Author: Guewen Baconnier
-#    Copyright 2012 Camptocamp SA
+#    Author: Anar Baghirli
+#    Copyright 2017 Camptocamp SA
```
yvaucher: 1. Don't change the author, you are not the original author of that module.
You can add your name as contributor in the README.rst
It's probably not what you meant to do but this is like stealing the work of someone else.
Add only your name as author if you are adding a functionnality or a major work in the code.

2. Dates shouldn't be erased, it should list the years of edition in this case 2012, 2017

3. Use the following format for headers:
https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L1-L3

yvaucher (changes requested):

yvaucher:
Please do a `git rebase -i` to clean up the commits

pingen/ir_attachment.py
```diff
@@ -1,123 +1,99 @@
...
-
         Returns None if the type is 'url' and the url is not reachable.
         """
         decoded_document = None
-        if attachment.type == 'binary':
-            decoded_document = base64.decodestring(attachment.datas)
-        elif attachment.type == 'url':
-            response = requests.get(attachment.url)
+        if self.type == 'binary':
+            decoded_document = self.datas
```
guewen: Can you explain why this should no longer be decoded? BTW the method name is `_decoded_content` so it isn't consistent.

abaghirli: it should be decoded. but not by base64.decodestring. that was the reason i got unreadable document.

abaghirli: here file is decoded:
https://github.com/OCA/report-print-send/pull/114/files#diff-ea036c00459d0303eb70c0c27049197eR148

PR #118, opened 2017-12-14
Porting to Odoo 11

ti-ku:
a quick approach making this module working in Odoo 11.


pedrobaeza:
There's already a migration in #110. Please check that one.

PR #122, opened 2018-01-11
[FIX] base_report_to_printer: Stateless tests

yajo:
If the test was run in a database with the `account` module installed, they were failing for 2 reasons:

- It was searching for the first report it could find, with the first 5 records to report. This made it load the `account.report_agedpartnerbalance` report, which caused a fake `TypeError: 'NoneType' object has no attribute '__getitem__'` error.
- It was running tests without loading the full module graph, thus not getting the default value for the new required `invoice_warn` field.

Now tests are run in `post_install` mode to load full module graph, and they use stateless data.

@Tecnativa

### DIFF
--- a/base_report_to_printer/tests/test_report.py
+++ b/base_report_to_printer/tests/test_report.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 LasLabs Inc.
+# Copyright 2017 Tecnativa - Jairo Llopis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import mock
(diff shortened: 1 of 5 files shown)


base_report_to_printer/tests/test_report.py
```diff
@@ -3,21 +3,43 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
pedrobaeza, line 4: Add your copyright

Garamotte (approved):
LGTM, thanks.

base_report_to_printer/tests/test_report.py
yajo: Done

pedrobaeza (approved):
We can merge this already as everything is green

PR #115, opened 2017-11-02
[9.0] Added graphic image support on ZPL labels

JosDeGraeve:
This PR adds support for graphic images on ZPL labels.

A new field type is defined which allow to print a static image or an image field.

TODO:
* update documentation

Depends on https://github.com/syleam/python-zpl2/pull/5

PR #129, opened 2018-01-14
[MIG][11.0] Migrated printer_zpl2 to v11.0

fmdl:
Standard migration
Add new features:
- Test Print Mode
- Test Emulation (from labelary.com)
- Button to add action in 'Action' Menu
- QRCode https://github.com/syleam/report-print-send/commit/8dd56b97b90f8762abbbcc4dc68eb06df36ce405
- Image support
- Import and convert ZPL2


fmdl:
cc @sylvain-garancher

pedrobaeza:
Travis is red

Garamotte:
@fmdl You forgot the `requirements.txt` file that allows Travis to install the `zpl2` python module.

Thanks for migrating the module.
I didn't look in details to your changes, but in case you didn't see it, there was already a PR to add the action in menu, maybe the contents or comments of this PR could be helpful here: https://github.com/OCA/report-print-send/pull/90
You may also be interrested to pick the graphics feature from this PR: https://github.com/OCA/report-print-send/pull/115

For the *test mode*, I think it would be better as a wizard than adding fields on the label which are not for the label itself.
Another customer already developped a preview using `http://api.labelary.com/v1/printers/{dpmm}/labels/{width}x{height}/{index}/{zpl}`, which generates a png file on the fly from ZPL code. Maybe we can use this ?

fmdl:
@sylvain-garancher @alnslang I think I have add every things.

Emulation change by `on_change`.

image


NOTA : there are bug a in odoo with the binary field, if you come back on a component with a picture, the on change fail. https://github.com/odoo/odoo/issues/22282

Garamotte:
Hello @fmdl 

Could it be possible to split the PR in two ?
One for the migration, another one for the improvements, based on the first.
This will allow to speed up merging the migration, then allow to review the improvements later, as they become big.

Also, I'm still not convinced by the "Test mode" print button, moreover with the preview from Labelary...

fmdl:
I am cannot split (I have not the history). Sébastien said me that the print button is a good idea. 

Florent de Labarre

Garamotte:
Anyway, it would be better to split a commit per improvement, instead of simply a commit for migration and a single commit for all improvements togegher, so you should to split your commits.

You can split a commit by using `git reset -p HEAD~`.
If this is not your last commit, then use `git rebase -i commit_to_split~` before, then select the `edit` action for this commit, and in the shell, use `git reset -p HEAD~` to split it and commit the split.

Sébastien said it is a good idea... I say this is not.
What's the benefit of adding some fields added on the `printing.label.zpl2` model are not related to a label, error-prone and not user-friendly, instead of using the already existing print wizard ?

fmdl:
Thanks for the git command. But I not a pro with command git, and I will not spend more time.

Anyway if you like you can disapprove and delete this PR (re-do from zero) but :
- I spent 2 days to migrate, add some great features (importation, emulation, test print, add tests on QR and Graphic).
- I add the button to quickly add action. like you say in the roadmap
- I add the labelary emulation like you said
- To add the labelary emulation I was obliged to add some "unrelated" field, 
- The labelary emulation are not excatly the reality, that why I was keep the test mode
- Me and other person use this module in my company, and I have developp a module to print test, because to use the wizard is so so long :
1- save the label
2- change tabs
3- click on action
4- click on Print label
5- select printer (2 click)
6- select label (2 click)
7- click Print
8- Return label tabs
Just by adding one more field, we can remove 8 clicks.

If you like to click, you can remove this feature, but I think the idea of Odoo is the simplicity. The main purpose to this module is to build easily Zebra Label, without any knowledge of ZPL2, why full testing is not builtin ?

Garamotte:
You said you didn't have the history, I explained how to split.
However, if the explanations are not enough, no problem, I did the split for you : https://github.com/syleam/report-print-send/tree/11.0-migrate-printer_zpl2-fmdl
(Edit: I ran the tests, and I fixed the commit of one test. Everything is OK now).

About the unit tests, you execute methods, but often don't check the results, so they are not valid.
Please add some assertions in tests.

Note that about the test mode, this is only my opinion, maybe all other people will like this feature.

fmdl:
@sylvain-garancher thanks for the split, it is more clean. I will add more test unit.

fmdl:
@sylvain-garancher I have add the last tests.

printer_zpl2/tests/test_generate_action.py
```diff
@@ -0,0 +1,42 @@
...
+            'model_id': self.env.ref(
+                'base_report_to_printer.model_printing_printer').id,
+        })
+
+    def test_create_action(self):
+        """ Check the creation of action """
+        self.label.create_action()
+        self.assertEqual(
+            1,
+            len(self.label.action_window_id))
```
Garamotte: To check if the record exists, you can simply use `self.assertTrue(self.label.action_window_id)`.

printer_zpl2/tests/test_generate_action.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_unlink_action(self):
+        """ Check the unlink of action """
+        self.label.unlink_action()
+        self.assertEqual(
+            0,
+            len(self.label.action_window_id))
```
Garamotte: Same as above, you can use `self.assertFalse(self.label.action_window_id)`.

printer_zpl2/tests/test_test_mode.py
```diff
@@ -0,0 +1,89 @@
...
+        self.label = self.env['printing.label.zpl2'].create({
+            'name': 'ZPL II Label',
+            'model_id': self.env.ref(
+                'base_report_to_printer.model_printing_printer').id,
+        })
+
+    def test_get_record(self):
+        """ Check if return a record """
+        self.label.record_id = 10
+        self.label._get_record()
```
Garamotte: Please add an assert here.

I think something like `self.assertEqual(self.label.get_record(), self.env[self.label.model_id.model].browse(10)` will work.

printer_zpl2/tests/test_test_mode.py
```diff
@@ -0,0 +1,89 @@
...
+        self.label.printer_id = self.printer
+        self.label.record_id = 10
+        self.label.print_test_label()
+        cups.Connection().printFile.assert_called_once()
+
+    def test_emulation_without_params(self):
+        """ Check if not execute next if not in this mode """
+        self.label.test_labelary_mode = False
+        self.label._on_change_labelary()
+        self.assertEqual(self.label.labelary_image, None)
```
Garamotte: Prefer using `self.assertIs(self.label.labelary_image, None)`.

printer_zpl2/tests/test_test_mode.py
```diff
@@ -0,0 +1,89 @@
...
+        self.label.test_labelary_mode = True
+        self.label.labelary_width = 80
+        self.label.labelary_height = 30
+        self.label.labelary_dpmm = '8dpmm'
+        self.env['printing.label.zpl2.component'].create({
+            'name': 'ZPL II Label',
+            'label_id': self.label.id,
+            'data': '"good_data"', })
+        self.label._on_change_labelary()
+        self.assertEqual(not(self.label.labelary_image), False)
```
Garamotte: Prefer using `self.assertTrue(self.label.labelary_image)`.

printer_zpl2/tests/test_wizard_import_zpl2.py
```diff
@@ -0,0 +1,94 @@
...
+        })
+        self.label = self.env['printing.label.zpl2'].create({
+            'name': 'ZPL II Label',
+            'model_id': self.env.ref(
+                'base_report_to_printer.model_printing_printer').id,
+        })
+
+    def test_open_wizard(self):
+        """ open wizard from label"""
+        self.label.import_zpl2()
```
Garamotte: Please add a simple assert to check the returned dictionnary (or at least some values like the model).

printer_zpl2/tests/test_printing_label_zpl2.py
```diff
@@ -0,0 +1,1061 @@
...
+        self.new_component({
+            'label_id': label.id,
+            'component_type': 'graphic',
+            'data': '"' + data + '"',
+            'height': 10,
+            'width': 10,
+            'reverse_print': 1,
+            'orientation': zpl2.ORIENTATION_ROTATED,
+        })
+        label._generate_zpl2_data(self.printer).decode("utf-8")
```
Garamotte: Please add an assert here.

printer_zpl2/tests/test_printing_label_zpl2.py
```diff
@@ -0,0 +1,1061 @@
...
+        """ Check contents of a image inverted label """
+        label = self.new_label()
+        data = 'R0lGODlhAQABAIAAAP7//wAAACH5BAAAAAAALAAAAAABAAEAAAICRAEAOw=='
+        self.new_component({
+            'label_id': label.id,
+            'component_type': 'graphic',
+            'data': '"' + data + '"',
+            'orientation': zpl2.ORIENTATION_INVERTED,
+        })
+        label._generate_zpl2_data(self.printer).decode("utf-8")
```
Garamotte: Please add an assert here.

printer_zpl2/tests/test_printing_label_zpl2.py
```diff
@@ -0,0 +1,1061 @@
...
+        """ Check contents of a image bottom label """
+        label = self.new_label()
+        data = 'R0lGODlhAQABAIAAAP7//wAAACH5BAAAAAAALAAAAAABAAEAAAICRAEAOw=='
+        self.new_component({
+            'label_id': label.id,
+            'component_type': 'graphic',
+            'data': '"' + data + '"',
+            'orientation': zpl2.ORIENTATION_BOTTOM_UP,
+        })
+        label._generate_zpl2_data(self.printer).decode("utf-8")
```
Garamotte: Please add an assert here.

Garamotte (commented):
Thanks.
Still some asserts missing.

I also added comments about the different assert methods, but this is not blocking if you don't change these.

max3903 (approved):

Garamotte:
Some fixes in https://github.com/syleam/report-print-send/commits/11.0-migrate-printer_zpl2-fmdl are still not applied here (mostly the raw option).

I fixed in the commits using rebase, and also added the diagonal lines management.

fmdl:
thank @sylvain-garancher , it is update.

printer_zpl2/wizard/wizard_import_zpl2.py
```diff
@@ -0,0 +1,437 @@
+import logging
+import re
+import base64
+import binascii
+import io
+
+
+from PIL import Image, ImageOps
+from odoo import models, api, fields, _
```
bodedra: remove `api,` its unused

printer_zpl2/models/printing_label_zpl2_component.py
```diff
@@ -0,0 +1,184 @@
...
+        selection=[
+            (zpl2.ORIENTATION_NORMAL, 'Normal'),
+            (zpl2.ORIENTATION_ROTATED, 'Rotated'),
+            (zpl2.ORIENTATION_INVERTED, 'Inverted'),
+            (zpl2.ORIENTATION_BOTTOM_UP, 'Read from Bottom up'),
+        ], required=True, default=zpl2.ORIENTATION_NORMAL,
+        help='Orientation of the barcode.')
+    diagonal_orientation = fields.Selection(
+        selection=[
+            (zpl2.DIAGONAL_ORIENTATION_LEFT, 'Left (\\)'),
```
bodedra, line 91: It fails with 

> AttributeError: module 'zpl2' has no attribute 'DIAGONAL_ORIENTATION_LEFT'

printer_zpl2/tests/test_generate_action.py
```diff
@@ -0,0 +1,38 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
bodedra, line 2: Please add copyright information. 

> Copyright <YEAR(S)> <AUTHOR(S)>

printer_zpl2/tests/test_printing_label_zpl2.py
```diff
@@ -0,0 +1,1118 @@
...
+            'model_id': self.env.ref(
+                'base_report_to_printer.model_printing_printer').id,
+        }
+        self.component_vals = {
+            'name': 'ZPL II Label Component',
+        }
+
+    def new_label(self, vals=None):
+        values = self.label_vals.copy()
+        if vals is not None:
```
bodedra: if vals:

printer_zpl2/tests/test_printing_label_zpl2.py
```diff
@@ -0,0 +1,1118 @@
...
+            values.update(vals)
+        return self.Model.create(values)
+
+    def new_component(self, vals=None):
+        values = self.component_vals.copy()
+        if vals is not None:
```
bodedra: same as above. `if vals`

printer_zpl2/tests/test_test_mode.py
```diff
@@ -0,0 +1,95 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
bodedra, line 2: Please add copyright information. 

> Copyright <YEAR(S)> <AUTHOR(S)>

printer_zpl2/tests/test_wizard_import_zpl2.py
```diff
@@ -0,0 +1,97 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
bodedra, line 2: Same missing copyright information.

printer_zpl2/views/printing_label_zpl2.xml
```diff
@@ -0,0 +1,199 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bodedra: Remove `encoding="UTF-8"` part. In python 3.x by default encoding is utf-8

printer_zpl2/wizard/print_record_label.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bodedra: Same remove `encoding="UTF-8"`

printer_zpl2/wizard/wizard_import_zpl2.py
```diff
@@ -0,0 +1,437 @@
...
+        if 'orientation' in vals.keys() and vals['orientation'] == '':
+            vals['orientation'] = 'N'
+
+        # Field
+        component = {}
+        for field, value in vals.items():
+            if field in self.model_fields.keys():
+                field_type = self.model_fields[field].get('type', False)
+                if field_type == 'boolean':
+                    if value == '' or value == zpl2.BOOL_NO:
```
bodedra: if not value or value == zpl2.BOOL_NO:

printer_zpl2/wizard/wizard_import_zpl2.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bodedra: Same remove `encoding="UTF-8"`

printer_zpl2/wizard/wizard_import_zpl2.xml
```diff
@@ -0,0 +1,24 @@
...
+                    <group>
+                        <field name="label_id"/>
+                        <field name="delete_component"/>
+                    </group>
+                </group>
+                <group string="ZPL2">
+                    <field name="data" widget="ace" nolabel="1"/>
+                </group>
+                <footer>
+                    <button type="special" special="cancel" string="Cancel"/>
```
bodedra: It would be good if we follow these new style in wizard button.

    <button string="Cancel" class="btn-default" special="cancel"/>
    <button string="Import" type="object" name="import_zpl2" class="btn-primary"/>

Reference: https://github.com/odoo/odoo/blob/11.0/addons/account/wizard/account_invoice_refund_view.xml#L43-L44

printer_zpl2/models/printing_label_zpl2_component.py
Garamotte: I'll check, maybe I didn't release the diagonal lines management in the zpl2 lib.

Garamotte: This should be OK now, I restarted the build on Travis.

max3903:
@fmdl Travis error and comment issues should be fixed by https://github.com/fmdl/report-print-send/pull/1

bodedra (approved):
LGTM. :+1:

max3903:
@sylvain-garancher @fmdl Are we ready to merge?

gurneyalex (approved):

Garamotte (approved):
Still not convinced by the "test mode", but if everyone is OK, I won't block :)

pedrobaeza:
Why not convinced, @sylvain-garancher ?

Garamotte:
@pedrobaeza I already explained before. In short:
- Both test modes: New fields added on the label that are not really related to the label itself, and not user-friendly (the `record_id` is a simple integer field), but I don't see a better way to do this, and the live preview is a great feature. This could be improved later.
- Auto-print only: I don't know how others work when creating labels. Personally, I save my changes very (very) often, but don't print a new label each time. There is already a generic label printing wizard, we could add a button to launch this wizard in case a test label is needed.

But this doesn't affect within the module's features, and we can disable this behaviour by a simple checkbox, so this is not a blocker for me. And as this is simply a test mode, it is very simple to change it later (no side-effect).

fmdl:
@sylvain-garancher there are no `Auto-print`, you need to click on the `Print` button to print :-)

printer_zpl2/models/printing_label_zpl2_component.py
fmdl: @sylvain-garancher I have update the view for the `diagonal` mode. :-)

Garamotte:
@fmdl Maybe I remember an old version then, thanks for pointing this out.

pedrobaeza:
Thanks for the explanations, @sylvain-garancher. Well, as you have explained, test mode seems harmless, but now Travis is red.

fmdl:
I will fix that this evenning

fmdl:
@pedrobaeza all lights are green

PR #130, opened 2018-01-14
[FIX] base_report_to_printer: <sheet> are not in correct place

fmdl:
And the views was ugly.


Garamotte (approved):
Thanks :)

pedrobaeza:
Please change module version number

yvaucher (approved):

yvaucher:
Will change version after merge

PR #132, opened 2018-01-22
[10.0][IMP] base_report_to_printer: Copies number in print report

luismontalba:
Add copies number to print report.

cc @Tecnativa

### DIFF
--- a/base_report_to_printer/models/printing_printer.py
+++ b/base_report_to_printer/models/printing_printer.py
@@ -5,6 +5,8 @@
 # Copyright (C) 2011 Domsense srl (<http://www.domsense.com>)
 # Copyright (C) 2013-2014 Camptocamp (<http://www.camptocamp.com>)
 # Copyright (C) 2016 SYLEAM (<http://www.syleam.fr>)
+# Copyright 2015 Oihane Crucelaegui - AvanzOSC
+# Copyright 2017 Luis M. Ontalba - Tecnativa
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import logging
@@ -94,7 +96,14 @@ class PrintingPrinter(models.Model):
             os.write(fd, content)
         finally:
             os.close(fd)
-
+        if copies == 1:
+            # If number of copies is not indicated by argument, check context
+            # or report definition
+            copies = (
+                self.env.context.get('report_copies') or
+                (report and report.report_copies) or
+                copies
+            )
         return self.print_file(
             file_name, report=report, copies=copies, format=format)
 
--- a/base_report_to_printer/models/printing_printer.py
+++ b/base_report_to_printer/models/printing_printer.py
@@ -73,6 +73,7 @@ class PrintingPrinter(models.Model):
         }
         return vals
 
+    # TODO Rename param report to report_name, to make behavior obvious
     @api.multi
     def print_options(self, report=None, format=None, copies=1):
         """ Hook to set print options """
@@ -83,8 +84,9 @@ class PrintingPrinter(models.Model):
             options['copies'] = str(copies)
         return options
 
+    # TODO Rename param report to report_name, to make behavior obvious
     @api.multi
-    def print_document(self, report, content, format, copies=1):
+    def print_document(self, report, content, format, copies=None):
         """ Print a file
 
         Format could be pdf, qweb-pdf, raw, ...
@@ -96,17 +98,19 @@ class PrintingPrinter(models.Model):
             os.write(fd, content)
         finally:
             os.close(fd)
-        if copies == 1:
+        report_obj = self.env["report"]._get_report_from_name(report)
+        if copies is None:
             # If number of copies is not indicated by argument, check context
             # or report definition
             copies = (
                 self.env.context.get('report_copies') or
-                (report and report.report_copies) or
-                copies
+                report_obj.report_copies or
+                1
             )
         return self.print_file(
             file_name, report=report, copies=copies, format=format)
 
+    # TODO Rename param report to report_name, to make behavior obvious
     @api.multi
     def print_file(self, file_name, report=None, copies=1, format=None):
         """ Print a file """
--- a/base_report_to_printer/tests/test_printing_printer.py
+++ b/base_report_to_printer/tests/test_printing_printer.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 LasLabs Inc.
+# Copyright 2017 Tecnativa.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 import tempfile
@@ -31,6 +32,9 @@ class TestPrintingPrinter(TransactionCase):
             'location': 'Location',
             'uri': 'URI',
         }
+        self.report = self.env['ir.actions.report.xml'].search([
+            ('report_type', '=', 'qweb-pdf'),
+        ], limit=1)
 
     def new_record(self):
         return self.Model.create(self.printer_vals)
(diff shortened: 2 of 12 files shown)


pedrobaeza:
Why that merge operation in commits? This is now red. Please check.

luismontalba:
Merge was due to solving a conflict in manifest file. 
Travis error seems to be related to other module.

base_report_to_printer/models/printing_printer.py
```diff
@@ -94,7 +96,8 @@ def print_document(self, report, content, format, copies=1):
             os.write(fd, content)
         finally:
             os.close(fd)
-
+        copies = (self.env.context.get('report_copies') or
```
pedrobaeza: Please put this:
```
copies = (
    self.env.context.get('report_copies') or
    (report and report.report_copies) or
    copies
)

**Review by pedrobaeza — CHANGES_REQUESTED** [review]
Travis error is not unrelated.

**Comment by luismontalba** [comment]
Changes done

**Review by pedrobaeza — APPROVED** [review]
@sergio-teruel please review

**Review by chienandalu — COMMENTED** [review]
Don't you test the copies stuff?

**Comment by pedrobaeza** [comment]
Well, not possible to test it I'm afraid.

**Review by chienandalu — APPROVED** [review]
Ok then :)

File: base_report_to_printer/tests/test_printing_printer.py [hunk 2]
```diff
@@ -55,7 +59,7 @@ def test_print_report(self, cups):
         with mock.patch('%s.mkstemp' % model) as mkstemp:
             mkstemp.return_value = fd, file_name
             printer = self.new_record()
-            printer.print_document('report_name', 'content to print', 'pdf')
+            printer.print_document(self.report, 'content to print', 'pdf')
```
pedrobaeza: @Yajo you changed this before, isn't it? Do you see this correct?

cubells (approved):
LGTM
Review code.

base_report_to_printer/tests/test_printing_printer.py
pedrobaeza: @Yajo, please confirm me this

yajo: I reviewed the code and I saw that this is now needed because [now it accesses a property from the report](https://github.com/OCA/report-print-send/pull/132/files#diff-b7f29d6ca081691b98c1e96480e7b69aR104). So, basically, we'd be changing the method signature now, expecting an ORM object instead of a string.

But to be honest, I can see that this argument wasn't used at all before, so I guess changing it wouldn't be such a big problem, unless some other addon relies on this one and expects to receive a string. Do you know if that could happen?

yajo (approved):

base_report_to_printer/tests/test_printing_printer.py
pedrobaeza: Well, I don't like too much that the argument is changed. Can you see if we can obtain the report from the report_name?

Garamotte: The test was "wrong" here.

The argument is already a `printing.printer` recordset everywhere in the code, examples:
https://github.com/OCA/report-print-send/blob/11.0/base_report_to_printer/models/ir_actions_report.py#L115
https://github.com/OCA/report-print-send/blob/11.0/base_report_to_printer/models/ir_actions_report.py#L146

Garamotte (approved):

base_report_to_printer/models/printing_printer.py
```diff
@@ -94,7 +96,14 @@ def print_document(self, report, content, format, copies=1):
             os.write(fd, content)
         finally:
             os.close(fd)
-
+        if copies == 1:
+            # If number of copies is not indicated by argument, check context
+            # or report definition
+            copies = (
+                self.env.context.get('report_copies') or
+                (report and report.report_copies) or
```
yajo: As commented in https://github.com/OCA/report-print-send/pull/132/files#r180006117, the `report` variable is not an ORM object, but a string with the report name.

This must be this way because [this method is called publicly from the XMLRPC interface](https://github.com/OCA/report-print-send/blob/2bae394d79bbab2508a1372ad8253db146b961ca/base_report_to_printer/static/src/js/qweb_action_manager.js#L26), which cannot send ORM objects.

Please get the report by other means, like searching by name.

base_report_to_printer/tests/test_printing_printer.py
yajo: You're right, this test must remain untouched.

yajo (changes requested):
Almost there pal! :wink:

base_report_to_printer/models/printing_printer.py
Garamotte: Oh, didn't see that the method was called through XMLRPC.
So we need to fix the methods that passes a recordset as argument in the `ir.actions.report` model...

pedrobaeza:
@Yajo as you know better this, please be you the one that fixes this.

yajo (approved):

Garamotte:
@Yajo You forgot to fix these calls:
https://github.com/OCA/report-print-send/blob/11.0/base_report_to_printer/models/ir_actions_report.py#L115
https://github.com/OCA/report-print-send/blob/11.0/base_report_to_printer/models/ir_actions_report.py#L146

yajo:
Those links are from branch 11. I think in branch 10 it is OK, isn't it?

pedrobaeza:
@Yajo any way, please remember to forward port this to v11 for not losing this on migration path.

yajo:
Shall we merge now?

pedrobaeza:
@sylvain-garancher you have the final call

Garamotte:
Oops, sorry, I did look in the wrong version.

Anyway, this seems wrong in v10.0 too, I think the passed argument should be `report_name` here:
- Line 23: [printer.print_document(report, document, report.report_type)](https://github.com/OCA/report-print-send/blob/10.0/base_report_to_printer/models/report.py#L23) where `report` is [self._get_report_from_name(report_name)](https://github.com/OCA/report-print-send/blob/10.0/base_report_to_printer/models/report.py#L16)
- Line 54: [printer.print_document(report, document, report.report_type)](https://github.com/OCA/report-print-send/blob/10.0/base_report_to_printer/models/report.py#L54) where `report` is [self._get_report_from_name(report_name)](https://github.com/OCA/report-print-send/blob/10.0/base_report_to_printer/models/report.py#L48)

yajo:
It should be ready now. I ammended to make it ready to merge.

I took the freedom to change the affected methods signatures to call the argument `report_name` instead of `report`, so the argument purpose becomes obvious.

Due to the signature change, I labeled this as `10.0.2.0.0`.

Garamotte:
I think renaming named arguments should be done only during major version migrations.
Travis fails because you forgot to rename the argument in the `print_option` method in `printer_tray`, but we also have custom modules using it. I'd prefer to keep the name `report` for existing versions, and maybe add a comment in the code to rename it for later versions.

yajo:
OK, I rolled back the signature changes and added some TODO notes.

Garamotte:
Thanks :+1: 

Last forgotten modification (give the name instead of the recordset) and Travis should be green: https://github.com/Tecnativa/report-print-send/blob/10.0-imp-base_report_to_printer/base_report_to_printer/tests/test_report.py#L107

yajo:
Thanks, let's see now. Sorry for force-pushing so much, I don't want to lose @luismontalba's commit

Garamotte (approved):
Everything is green, you can merge now :)

pedrobaeza:
@luismontalba please port this to v11.

PR #136, opened 2018-03-06
[IMP] base_report_to_printer: existing printers migration

chienandalu:
- Avoid duplicating printers when `action_update_printers()` is hit for the first time coming from a previous version.

cc @Tecnativa


pedrobaeza (approved):
Fast-tracking as this is a migration script

Issue #134, opened 2018-02-08
[11.0] Unable to generate a report without being member of Print/User group


Garamotte:
When a user wants to generate a report, the `_get_user_default_print_behaviour` method searches for the default printer. This operation needs at least Printer/User access rights, so we have to give this group to all employees.
In addition, this error is silenced, and the traceback is not available server-side, it's only displayed in the browser's debug console.

I think that we should check the access rights before searching for a printer.

pedrobaeza:
Or maybe with `sudo`? Which version?

Garamotte:
I had the issue on version 11.0, don't know for other versions.

pedrobaeza:
Well, you know better than me the module, so you can take the best technical decision 😉

schout-it:
Only thing I did not test yet, is automatic sale confirmation of website sales and portal users.
do not now it they also need access to these objects or if there should be a different flow in those cases
(would expect that for portal users, you almost always send to client)

PR #138, opened 2018-03-14
[9.0][10.0][11.0] [FIX] user access to printer objects

schout-it:
#134 
This was default behavior in v8 where there was no specific group on the read access of these objects.
The group was created in v9 so all versions upwards are affected.

### DIFF
--- a/base_report_to_printer/__manifest__.py
+++ b/base_report_to_printer/__manifest__.py
@@ -8,7 +8,7 @@
 
 {
     'name': "Report to printer",
-    'version': '10.0.1.0.2',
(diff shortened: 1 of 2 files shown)


Garamotte (approved):
LGTM, thanks.

pedrobaeza (approved):
Please change the module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

schout-it:
version changed to 10.0.1.0.3

pedrobaeza:
The same should be needed for the rest of the versions.

PR #141, opened 2018-04-12
[10.0][FIX] base_report_to_printer: Loss report_copies context

sergio-teruel:
In this call the report_copies context is lost
cc @Tecnativa


carlosdauden (approved):

sergio-teruel:
@pedrobaeza module version changed

pedrobaeza (approved):
We can fast-track when CI finishes

PR #143, opened 2018-04-18
[FIX] printer_zpl2: don't take the bin_size during the onchange

fmdl:
It fix an issue when you use the Labelary Mode if case you have change a component with an image.

fix an issue of the framework : https://github.com/odoo/odoo/issues/22282#issuecomment-382372340

cc @sylvain-garancher


fmdl:
@alnslang

fmdl:
fix in V12

PR #144, opened 2018-04-19
[FIX] printer_zpl2: record_id is not set in some case

fmdl:
fix : record_id is not set in some case

@sylvain-garancher


fmdl:
@alnslang

fmdl:
fix in v12

PR #146, opened 2018-05-03
[ADD] label_zpl2 and make printer_zpl2 depend to it

etobella:
I have detected that `printer_zpl2` depends on `base_report_to_printer` but I think it may not be the best options, because you could require zpl2 but not use the module of printers.

This PR splits the module in two. The first is `label_zpl2`, that creates the models and allows to render the labels. The second is `printer_zpl2` that will inherit from the first.

### DIFF
--- a/label_zpl2/README.rst
+++ b/label_zpl2/README.rst
@@ -2,20 +2,27 @@
    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
    :alt: License: AGPL-3
 
-=====================
-ZPL II Label printing
-=====================
+============
+ZPL II Label
+============
 
-This module extends the **Report to printer** (``base_report_to_printer``)
-module to add a ZPL II label printing feature.
+This module adds the possibility to define printing labels in the
+Zebra Programming Language (ZPL II).
 
-This module is meant to be used as a base for module development, and does not provide a GUI on its own.
-See below for more details.
+See https://en.wikipedia.org/wiki/Zebra_(programming_language)
+
+This module is meant to be used as a base for module development, and does
+not provide a GUI on its own. See below for more details.
 
 Installation
 ============
 
-Nothing special, just install the module.
+Install python module 'zpl2' - https://pypi.org/project/zpl2
+
+.. code-block:: bash
+
+   sudo pip install zpl2
+
 
 Configuration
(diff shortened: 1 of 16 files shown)


JordiBForgeFlow:
@sylvain-garancher As the original author of this module, it would be best if you can tell us if you're good with this proposal. Can you help us?

max3903:
@etobella I am not sure to understand why you would split. If you create ZPL2 labels, I would assume it's to print them on a ZPL2 printers, no? Can you explain your use case? Do you plan to use one module and not the other?

Garamotte:
Like @max3903, I don't understand the usecase, but I'm not against splitting the module if it contains several independant features, as long as everything works after the split :)

However, I think that the split should wait for the next Odoo version.
Some people may have modules depending on this version, and your split changes many things (xml_ids of the models, views, menus, etc.) that can break custom modules inheriting from this one.

etobella:
Well, I am using odoo integrated with another software and all the printing queres are managed from the other software.
All data is stored on odoo, so I tried to generate the labels on odoo and use the printing queues from the other software. I am currently using it in order to print complex templates and I would like to use it also for labels.

max3903:
@etobella Thanks for the explanation. It makes more sense now.

I don't mind merging this in v11 as far as the version is bumped accordingly.

max3903:
@etobella Can you split the pull request to have one per module? Thanks.

etobella:
I have divided the module in two. (#150)

max3903 (approved):

JordiBForgeFlow:
@sylvain-garancher @alnslang The module was migrated by you to 11.0 just two months ago. I think that it is important to coordinate the changes with you. But I do not expect many other people to use the modules. The module is not yet available in the app store for 11.0 https://www.odoo.com/apps/modules/10.0/printer_zpl2/

Also, if a person fetched all the repo the update will run correctly, and will install label_zpl2

pedrobaeza:
More than to use it, to extend it, so the changes on XML-IDs won't affect. I think this can be very valuable for these regular contributors in this version (they can't wait for next version), and with a very little cost.

etobella:
I cancel the PR as it will not be merged for the related problems found for migration.

PR #148, opened 2018-05-07
[FIX] Access denied to employee

bodedra:


pedrobaeza:
I don't think this should be so opened. Please restrict a bit more. Any way, is there any use case when you have an error? If so, please include steps to reproduce.

bodedra:
@pedrobaeza In our case, Some users do not have access to printer rights. They try to print it and got error of access denied.

CC @max3903 @guewen @jgrandguillaume

pedrobaeza:
But using regularly?

base_report_to_printer/views/printing_printer.xml
```diff
@@ -4,7 +4,7 @@
     <menuitem name="Printing"
         id="printing_menu" sequence="1"
         parent="base.menu_administration"
-        groups="printing_group_manager"
+        groups="printing_group_manager,base.group_user"
```
max3903: @bodedra I don't think `group_user` can access the Settings app...

max3903:
@bodedra What is the object in the access denied message?

max3903 (approved):

yvaucher (approved):
LGTM, It follows the same logic than base modules

PR #154, opened 2018-06-21
[FIX][base_report_to_printer] - print documents with ir.actions.repor…

ougc27:
### Impacted Versions
11.0

### Steps to Reproduce

1. Configure a CUPS server in Odoo.

2. Search the active printers

3. Configure the report action (send_to_printer) and select the printer.

4. Print the report.

### Current Behaviour
The module returns an error ( "Error when sending the document to the printer").
This error is launched because the module when a report is sended to the printer from the method "render_qweb_pdf", self is an instance of ir.actions.report without any object and on the super call Odoo try to instance the active model with self.active_model and returns a traceback with a KeyError.

captura realizada el 2018-06-21 09 29 48

###  Desired Behaviour
The report must be sendend to the printer successfuly.

### Extra
I attach a video with the behaviour.
https://youtu.be/3GU1Q-sjtl8

@etobella could you help me reviewing this PR please? :pray: 
Regards

### DIFF
--- a/base_report_to_printer/models/ir_actions_report.py
+++ b/base_report_to_printer/models/ir_actions_report.py
@@ -129,6 +129,13 @@ class IrActionsReport(models.Model):
             return True
         return False
 
+    @api.noguess
+    def report_action(self, docids, data=None, config=True):
+        res = super().report_action(docids, data=data, config=config)
+        if not res.get('id'):
+            res['id'] = self.id
+        return res
+
     def render_qweb_pdf(self, docids, data=None):
         """ Generate a PDF and returns it.
 
--- a/base_report_to_printer/__manifest__.py
+++ b/base_report_to_printer/__manifest__.py
@@ -7,7 +7,7 @@
 
 {
     'name': "Report to printer",
-    'version': '11.0.2.2.0',
+    'version': '11.0.2.3.0',
     'category': 'Generic Modules/Base',
     'author': "Agile Business Group & Domsense, Pegueroles SCP, NaN,"
               " LasLabs, Camptocamp, Odoo Community Association (OCA)",


pedrobaeza:
You are modifying the method signature, which is not valid for stable versions. Can you do it other way? @sylvain-garancher what do you think about this?

etobella:
I found the issue. It is related to the function report_action, as the api is expecting a id or active_ids, but it is returning none.
You should add on `ir_action_report` the following code:
    @api.noguess
    def report_action(self, docids, data=None, config=True):
        res = super().report_action(docids, data=data, config=config)
        if not res.get('id', False):
            res['id'] = self.sudo().id
        return res

etobella:
Or do you prefer to close this PR and I make it?

ougc27:
Hello @etobella 

I made the changes that you recommended me, please review it.
I wait for your feedback.

Best Regards!

etobella:
With my change it was enough. Could you try to remove your first commit and leave only my proposal?

First, you could check that it works as expected. I tried it locally and it looked good on my site, but it is allways better to have an extra pair of eyes checking it.

ougc27:
Hello again @etobella 

I already remove the previous commit, I tried locally too with success results. I think that works with any problem.

etobella:
Perfect!
As it was a modification that you and me agreed on, I think I should not approve this PR, so we should need to wait for someone to check it :wink:

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -129,6 +129,13 @@ def _can_print_report(self, behaviour, printer, document):
             return True
         return False
 
+    @api.noguess
+    def report_action(self, docids, data=None, config=True):
+        res = super().report_action(docids, data=data, config=config)
+        if not res.get('id', False):
```
Garamotte: I think `False` is not required here, because the default value is `None`, which is evaluted to `False` in an `if` statement.

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -129,6 +129,13 @@ def _can_print_report(self, behaviour, printer, document):
...
+            res['id'] = self.sudo().id
```
Garamotte: Why `sudo` ?

pedrobaeza:
@sylvain-garancher what do you think about this? At least you should bump version number

base_report_to_printer/models/ir_actions_report.py
ougc27: You're right, Done!. :+1:

base_report_to_printer/models/ir_actions_report.py
ougc27: Done! :+1:

Garamotte (approved):

PR #156, opened 2018-06-28
[FIX] Group inheritance

max3903:
* "Print Manager" implies "Print User"
* "Employee" implies "Print User"
* "Admin / Access Rights" implies "Print Manager"

### DIFF
--- a/base_report_to_printer/security/security.xml
+++ b/base_report_to_printer/security/security.xml
@@ -1,116 +1,126 @@
 <?xml version="1.0" ?>
-<odoo noupdate="1">
-    <data>
-        <record id="printing_group_manager" model="res.groups">
-            <field name="name">Printing / Print Manager</field>
-            <field name="users" eval="[(4, ref('base.user_root'))]"/>
-        </record>
-        <record id="printing_group_user" model="res.groups">
-            <field name="name">Printing / Print User</field>
-            <field name="users" eval="[(4, ref('base.group_user'))]"/>
-        </record>
-        <record id="printing_server_group_manager" model="ir.model.access">
-            <field name="name">Printing Server Manager</field>
-            <field name="model_id" ref="model_printing_server"/>
-            <field name="group_id" ref="printing_group_manager"/>
-            <field eval="1" name="perm_read"/>
-            <field eval="1" name="perm_unlink"/>
-            <field eval="1" name="perm_write"/>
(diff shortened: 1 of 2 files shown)


bodedra (approved):
LGTM.

etobella (approved):
LGTM. Version should be changed?

pedrobaeza:
As you have changed all the indentation levels, it's very difficult to follow changes. Please keep the indentation (although you remove one XML level).

Garamotte:
@pedrobaeza `git diff` has a `-b` parameter to hide withespace changes.
GitHub now has this feature too
selection_015

pedrobaeza:
Great, I didn't know about that option

base_report_to_printer/security/security.xml
```diff
@@ -1,116 +1,126 @@
 <?xml version="1.0" ?>
 <odoo noupdate="1">
```
pedrobaeza: Remove noupdate="1"

Garamotte (approved):

pedrobaeza (changes requested):
And please change module version number

base_report_to_printer/security/security.xml
Garamotte: I agree, but this will not change on already installed databases.
I suggest to change the noupdate flag with a pre-migration script if we decide to remove noupdate here.

pedrobaeza: There was a `<data>` tag after that has been removed now, so the final value was `noupdate="0"`, and thus no need of migration script:

seleccion_017

Garamotte: Oh, ok, I never encountered a `<data>` tag without noupdate in a `<odoo noupdate="1">` tag before. We learn a lot in this PR :)

bodedra: If we remove `noupdate=1` then might be chance to lost updated value when we upgrade module in same database.  :confused: For example: directly update value from browser.

pedrobaeza: No, as said, this has no effect, as there was after a `<data>` tag. Now that you are removing that tag, you have to remove the `noupdate` for letting the same situation.

max3903:
@pedrobaeza Version bumped.

Thank you all!

pedrobaeza (approved):

PR #157, opened 2018-07-03
[9.0][IMP] Change version to allow install in Ubuntu 16.04+

igallart:
The version 1.9.66 is very old, and is not possible install in Ubuntu 16.04. Presumably, in future versions either.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pycups==1.9.66
+pycups>=1.9.68
 PyPDF2==1.18
 requests
 zpl2


pedrobaeza:
We told about not blocking versions, or at least put `>=`, isn't it, @sylvain-garancher ?

igallart:
I do not know if higher versions will work. For me no problem.

sbidoul:
Yes please no pinned versions in requirements.txt. Use `>=` if you know the minimum required version.

Garamotte:
I don't know why this module is pinned to version 1.9.66, but I can confirm that versions below 1.9.68 cannot be installed on Ubuntu 16.04, so `pycups>=1.9.68` seems good, as [we shouldn't pin versions in `requirements.txt`](https://github.com/OCA/maintainer-quality-tools/issues/494).

I personnally always use the latest version without any issue (which is also very old).

yvaucher (approved):

PR #159, opened 2018-07-12
[8.0][IMP] Change version to allow install in Ubuntu 16.04+

igallart:
Same problem that https://github.com/OCA/report-print-send/pull/157/files but version 8.0


pedrobaeza:
Please put `>=` as well.

yvaucher:
@pedrobaeza it seems it already `>=` in here

yvaucher (approved):

pedrobaeza (approved):

PR #160, opened 2018-07-30
[10.0][FIX] print_options expecting the full report instead of a string

ecino:
When trying to print report with printer_tray module installed, commit 3eb0c70acbc7105c6bb740fb4cd1b7beb60f599c changed how parameters are send and this causes a bug on printer_tray module which was expecting a record parameter instead of a string. We propose to check first if we receive a report name as string.


ecino:
I think this is a bug fix that should be quickly reviewed and merge because we are seeing errors using the current version.

pedrobaeza (approved):
OK, fast-tracking this.

PR #164, opened 2018-11-01
Add @api.model decorator on create, write

nbessi:
on pingen ir.attachment.

When multiple modules redefine create, write it can lead to TypeError

### DIFF
--- a/pingen/ir_attachment.py
+++ b/pingen/ir_attachment.py
@@ -23,7 +23,7 @@ import requests
 import base64
 
 from openerp.osv import osv, orm, fields
-from openerp.tools.translate import _
+from openerp.tools.translate import _, api
 
 
 class ir_attachment(orm.Model):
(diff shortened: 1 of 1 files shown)


pingen/ir_attachment.py
```diff
@@ -92,12 +92,14 @@ def _handle_pingen_document(self, cr, uid, attachment_id, context=None):
...
         return
 
+    @api.model
     def create(self, cr, uid, vals, context=None):
         attachment_id = super(ir_attachment, self).create(cr, uid, vals, context=context)
         if 'send_to_pingen' in vals:
             self._handle_pingen_document(cr, uid, attachment_id, context=context)
         return attachment_id
 
+    @api.model
```
pedrobaeza, line 102: Write shouldn't be `@api.model`.

gdgellatly: Moreover, if you are adding the decorator, you better change the function signature.

dreispt, line 102:
```suggestion
```

dreispt (commented):
Good to go one `write()` fix is applied

PR #165, opened 2018-11-07
[11.0][remote_report_to_printer] add permissions to write and delete to base.group_system

JordiBForgeFlow:
Otherwise only the admin user was able to edit / delete remote printers.


etobella (approved):
LGTM :+1: 
Version should be changed?

PR #169, opened 2019-01-30
12.0 mig base report to printer

gdgellatly:
Confirmed working
- printing directly to printer from report action
- updating printers from CUPs
- setting up a report to print directly
- menus and views
- all tests passing
- updating printer trays from CUPS
- updating jobs w/ bug fixes (new security rule, and allowance for non Odoo printers)

Question for JS type people:
I've moved the intercept of the report action up into `_triggerDownload` and essentially recreated it there. At the moment I've also removed the pdf checking, as I want to test with other report types. 

However I have a feeling we could massively simplify this even further but don't know how.  Currently the behaviour without the intercepting javascript is to perform a print and also the client action (`_downloadReport`). I feel like if we can just intercept that return to block that download it might be easier.


fmdl:
@gdgellatly thank for the job, can you fix that:
`base_report_to_printer/__manifest__.py:13:80: E501 line too long (100 > 79 characters)`

it is a final version to test ?

gdgellatly:
@fmdl yes, its in production 6 weeks. The only issues I've really seen so far is that 

- certain HP printers don't have InputSlot in their PPD's so trays don't populate, but can be specified manually in print options as HPPaperSource. 
- Also where printers have multiple tray options, but not installed it brings back all of them.

But I guess that was an issue before anyway.

fmdl (approved):

fmdl:
@sle-subteno-it

fmdl:
@sla-subteno-it

fmdl:
@gdgellatly can you sqaush your migration commit (and remove the WIP) ? :-)

AdriaGForgeFlow (approved):
Code review :+1:

gdgellatly:
I'll do a rebase over the weekend for this

pedrobaeza:
OK, thanks. Please squash migration commits at your taste.

sla-subteno-it (approved):

PR #172, opened 2019-03-12
[FIX] printer_zpl2: QR as automatic input

etobella:
Without this PR, QR can only be alphanumeric (upper case) and now, it will detect automatically what to do. For example, now, you can insert a json inside a QR.

### DIFF
--- a/printer_zpl2/__manifest__.py
+++ b/printer_zpl2/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'Printer ZPL II',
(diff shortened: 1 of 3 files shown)


jarroyomorales (approved):
LGTM :+1:

JordiBForgeFlow (approved):

JordiBForgeFlow:
@etobella If you can squash the commits I'll then merge.

etobella:
@jbeficent Squashed!

PR #173, opened 2019-03-20
[MIG] printer_zpl2: Migration to 12.0

fmdl:
@sla-subteno-it

PR #174, opened 2019-04-19
[MIG] base_report_to_printer and printer_zpl2 migration v12

jco-subteno-it:


pedrobaeza:
There's already #169 and #173, and yours even doesn't respect commit history, so please review that one instead of opening a new one.

PR #189, opened 2019-10-25
[FIX] .travis.yml,requirements.txt: Move python deps to the right place

luisg123v:
Currently, some Python dependencies are being installed manually from the `.travis.yml` file, using
pip install python_package

Not only that way is a bad practice, but also prevents those Python modules to be retrieved when an Odoo module is installed via MQT.

In addition, the modules `requests` and `PyPDF2` are removed, as they already are in Odoo's requirements.


luisg123v:
@moylop260
Should I close this PR?

I just noticed that cups
 is actually being installed, but via the `.travis.yml`.

moylop260:
Its better using requirements.txt file in order to propagate it
So, you can delete it from .travis.yml

moylop260:
in fact, move all “pip install” commands from .travis.yml to requirements.txt file, please

requirements.txt
```diff
@@ -1 +1,4 @@
 zpl2>=1.1
+pycups==1.9.73
+PyPDF2==1.18
+requests
```
flotho: IMHO, it should be great if you specify a version here.
Odoo core is having its version requests https://github.com/odoo/odoo/blob/12.0/requirements.txt#L40 and your module could override it in an unwanted manner.

Regards

flotho (changes requested):

requirements.txt
luisg123v: @flotho so, if it's already in Odoo's requirements, IMHO it should be removed from here. It makes no sense to re-define it from here. Don't you think?

The same applies for PyPDF2, where the version specified in Odoo is even greater.

sbidoul: Do we really need to pin requirements here? It's preferable to test with the latest version of the the dependencies, and specify a minimum required version in `requirements.txt`.

luisg123v:
@moylop260 @flotho @sbidoul 
This is ready for review.

What I did:
- Move Python dependencies from the `.travis.yml` file to the `requirements.txt` file
- Remove dependencies that are already specified in Odoo's requirements
- Update commit/PR description accordingly

flotho (approved):

PR #190, opened 2019-11-28
[11.0][ADD] Stock picking auto print

bodedra:
This module will send default picking report to printer when delivery order is ready or done. It will find default picking report based on following configuration.

Configuration
=============

* Active developer mode. Go to Technical > Actions > Reports
* Edit report with following fields.

   * Is Default Report?
   * Country and
   * Company


pedrobaeza:
Please don't remove `stock_` prefix in the module name for proper main area recognition.

bodedra:
@pedrobaeza Make sense. :+1:

bodedra:
Are you available for review it?

@b-kannan @dreispt @max3903  @gavindav @diverfr @mdodoo @guimarc-br

stock_picking_auto_print/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    'name': 'Direct Print',
+    'summary': 'Auto print when DO is ready',
+    'version': '11.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Open Source Integrators, Odoo Community Association (OCA)',
+    'category': 'Generic Modules/Base',
+    'website': 'http://www.opensourceintegrators.com',
+    'depends': [
+        'stock',
```
max3903:
```suggestion
```

stock_picking_auto_print/models/ir_actions_report.py
```diff
@@ -0,0 +1,13 @@
...
+# Copyright (C) 2019 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    is_default_report = fields.Boolean('Is Default Report?')
```
max3903, line 11:
```suggestion
    is_default_report = fields.Boolean('Is Default Report?', default=False)
```

max3903: @bodedra I am guessing there should only be one default per country or company, no?

bodedra: IMO, Boolean is by default is false. Ref https://github.com/odoo/odoo/blob/11.0/odoo/models.py#L3478

Yes. We can add constraint for one default per country AND company. Regarding per country OR company, may we leave it to decide User? Currently, this module will fetch only one records set. Or any other method to restrict except create()/write() method?

stock_picking_auto_print/models/stock_picking.py
```diff
@@ -0,0 +1,84 @@
...
+                    limit=1)
+                if report_ids:
+                    default_report_id = report_ids.id
+
+            if not default_report_id:
+                default_report_id = self.env.ref(
+                    'stock.action_report_picking').with_context(
+                    landscape=True).report_action(picking).get('id')
+            action_report = report_action_pool.browse(
+                default_report_id)
```
max3903, line 71: @bodedra Why not using search_read?

```
search_read(['company_id', 'country_id'], [('model', '=', 'stock.picking'), ('is_default_report', '=', True)])
```

and then use the results to determine which one to use.

bodedra: @max3903 I like the idea. But there is sequential order given to fetch default report. Here is sequence:

1. Country and Company
2. Country,
3. Company.

If not found, call "Picking Operations" report.

Is there way, we can identify our need using "search_read"?

max3903: @bodedra Yes, instead of asking the DB and potentially running up to 4 SQL queries, use the list of dictionary returned by search_read:

```
[{'id': 1, 'company_id': 2, 'country_id': 3},
{'id': 4, 'company_id': 5, 'country_id': False},
{'id': 6, 'company_id': False, 'country_id': 7},
{'id': 8, 'company_id': False, 'country_id': False}]
```

max3903:
```
res = ...search_read(...)
for row in res:
    if row['company_id] == ... and row['country_id'] == ...:
        ...
    elseif row['country_id'] == ...:
        ...
    elseif row['company_id'] == ...:
        ..
    else:
        ...
browse
```

bodedra: @max3903 Make sense. :+1:

bodedra:
@max3903 Comment addressed!

bodedra:
Ping @b-kannan @dreispt  @gavindav @diverfr @mdodoo @guimarc-br

max3903 (approved):

dreispt (approved):

PR #195, opened 2019-12-23
[FIX] base_report_to_printer: Migrate noupdate record

yajo:
In version 9.0.2.0.0, added in https://github.com/OCA/report-print-send/pull/60, the cron was modified. However, since it is a noupdate=1 record, preinstalled instances started failing there.

This fixes it.

@Tecnativa TT18838

### DIFF
--- a/base_report_to_printer/migrations/9.0.2.0.0/post-10-create_server_record.py
+++ b/base_report_to_printer/migrations/9.0.2.0.0/post-10-create_server_record.py
@@ -18,3 +18,13 @@ def migrate(cr, v):
             'port': config.get('cups_port', 631),
             'printer_ids': [(6, 0, env['printing.printer'].search([]).ids)],
         })
+        # Update the noupdate=1 cron if modified fields weren't touched
+        cron = env.ref("base_report_to_printer.ir_cron_update_printers")
+        if (
+            cron.model == "printing.printer" and
+            cron.function == "update_printers_status"
+        ):
+            cron.write({
+                "function": "action_update_jobs",
+                "model": "printing.server",
+            })


pedrobaeza:
This is done through a file (which we call it `noupdate_changes.xml`) and a call to an openupgradelib method. The file is obtained with this script: https://github.com/OCA/OpenUpgrade/blob/12.0/scripts/compare_noupdate_xml_records.py and check an example of its usage: https://github.com/OCA/OpenUpgrade/blob/bdf26ceb11834fd45ca4b4577532d6f5ee2ca4e4/addons/hr/migrations/12.0.1.1/post-migration.py#L22

yajo:
Do I really must do all of that? :roll_eyes: This has nothing to do with openupgrade...

pedrobaeza:
It's not a question of OpenUpgrade, but migration stuff, and it's a way to don't add so much code: you standarize the way to obtain noupdate records changes and the call you do for loading that stuff. I know this time seems more work, but doing this way you won't pass any other change unaware.

yajo:
It seems to not work:

```
doodba INFO: Executing python custom/src/odoo/scripts/compare_noupdate_xml_records.py custom/src/report-print-send/base_report_to_printer_old custom/src/report-print-send/base_report_to_printer_old --mode module


base_report_to_printer_old:

<?xml version='1.0' encoding='utf-8'?>
<openerp>
  <data/>
</openerp>
```

Am I missing something?

yajo:
I updated travis because it seemed it was failing because it was outdated. However now it fails with a timeout. Any clues?

pedrobaeza:
Maybe the changes are in other version? I'm now on mobile and I can't check more, sorry.

yajo:
I copied the module in another folder in the version where the changes were present.

rafaelbn:
Should we maintain this old PR open? Is v9 maybe we could close it @Yajo

pedrobaeza:
Yes, closing for now.

PR #213, opened 2020-04-24
[9.0][FIX] base_report_to_printer: Fix api.v8 get_pdf signature

grindtildeath:
I'm a bit confused with the change done on #77 for api.v8 `get_pdf`: https://github.com/OCA/report-print-send/pull/77/files#diff-0729f9ddc4d839c5661dad89db3bb9b5R59-R62

On OCA/OCB as well as on odoo/odoo, this method expects a recordset:
 - https://github.com/OCA/OCB/blob/9.0/addons/report/models/report.py#L272-L275
 - https://github.com/odoo/odoo/blob/9.0/addons/report/models/report.py#L272-L275

I guess it did work as long as no other module redefines this function, but then it's broken if one uses other modules as it the case with: https://github.com/OCA/l10n-switzerland/blob/9.0/l10n_ch_payment_slip/report/payment_slip_from_invoice.py#L72-L77

ping @sylvain-garancher @yvaucher @guewen

### DIFF
--- a/base_report_to_printer/models/report.py
+++ b/base_report_to_printer/models/report.py
@@ -37,29 +37,21 @@ class Report(models.Model):
             return True
         return False
 
-    @api.v7
-    def get_pdf(self, cr, uid, ids, report_name, html=None,
-                data=None, context=None):
+    @api.model
+    def get_pdf(self, docids, report_name, html=None, data=None):
         """ Generate a PDF and returns it.
 
         If the action configured on the report is server, it prints the
         generated document as well.
         """
-        document = super(Report, self).get_pdf(cr, uid, ids, report_name,
-                                               html=html, data=data,
-                                               context=context)
-        report = self._get_report_from_name(cr, uid, report_name)
+        report = self._get_report_from_name(report_name)
+        docs = self.env[report.model].browse(docids)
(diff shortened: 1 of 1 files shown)


bealdav:
Probably the same problem ?

solved here

https://github.com/OCA/edi/pull/194/files

yvaucher:
@bealdav Thanks @alexis-via solution seams a bit more elegant to show explicitely that we use an `@api.model`

@grindtildeath can you try by using `@api.model` ?

i-vyshnevska (approved):
LG, 
I have seen similar travis failed to the build in other repos it not related to the change

grindtildeath:
@bealdav Thanks for the hint :wink: 
@yvaucher Using `api.model` on `get_pdf` requires an extra `browse` in this method to fetch the records, only to get their ids in the next call to `get_pdf` as it will be on `api.v8` after the call to `super`... 
Therefore I'm not really sure it's much better than using `api.v7` and `api.v8` decorators :shrug:

yvaucher:
@grindtildeath thanks for the change, I think we can live with the extra browse. It's less confusing that way.

yvaucher (approved):

sebalix (approved):

PR #214, opened 2020-04-27
[9.0][ADD] base_report_to_printer: Print attachments wizard

grindtildeath:


base_report_to_printer/wizards/print_attachment_report.py
```diff
@@ -0,0 +1,56 @@
...
+            self.printer_id.print_document(
+                None,
+                content=content,
+                format=content_format,
+                copies=att_line.copies
+            )
+        self.attachment_line_ids.unlink()
+
+
+class PrintAttachment(models.TransientModel):
```
i-vyshnevska:
```suggestion
class PrintAttachmentLine(models.TransientModel):
```
same name class

base_report_to_printer/wizards/print_attachment_report.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2020 Camptocamp SA
```
sebalix, line 2: `# -*- coding: utf-8 -*-` as we are running on Python 2

base_report_to_printer/wizards/print_attachment_report.py
```diff
@@ -0,0 +1,56 @@
...
+        'wizard.print.attachment.line', 'wizard_id',
+        string='Attachments to print',
+    )
+
+    @api.multi
+    def print_attachments(self):
+        """ Prints a label per selected record """
+        self.ensure_one()
+        for att_line in self.attachment_line_ids:
+            content = base64.b64decode(att_line.attachment_id.datas)
```
sebalix: Should we test that `attachment_id.datas` could return `False` (not required on `ir.attachment`)?

base_report_to_printer/wizards/print_attachment_report.py
```diff
@@ -0,0 +1,56 @@
...
+        self.attachment_line_ids.unlink()
+
+
+class PrintAttachmentLine(models.TransientModel):
+    _name = 'wizard.print.attachment.line'
+    _description = 'Print Attachment line'
+
+    wizard_id = fields.Many2one("wizard.print.attachment")
+    attachment_id = fields.Many2one(
+        'ir.attachment',
```
sebalix, line 60: `required=True`?

sebalix (commented):
Small comments

base_report_to_printer/wizards/print_attachment_report.py
grindtildeath: Not sure it will happen, it's not required because it's computed using `store_fname` or `db_datas` but I added a warning message anyway. :wink:

sebalix (approved):

yvaucher:
@grindtildeath a description of the change would be nice.

PR #220, opened 2020-07-15
[12.0][ADD] base_report_to_printer: Print attachments wizard

KARIZMA-CONSEIL:

PR #222, opened 2020-08-06
[12.0][ADD] base_report_to_printer: Print attachments wizard

sebalix:
Port of #214 (supersedes #220)


grindtildeath (approved):
I guess we can keep this one and close both #220 and #223

yvaucher:
@sebalix can you describe what does the change, there is no description in #214

gurneyalex (approved):

PR #225, opened 2020-08-24
[12.0] ADD printer_zpl2_browser_print

eLBati:
Allow to print ZPL labels directly from web browser, using printers installed in client device.

See [Zebra Browser Print](https://www.zebra.com/us/en/products/software/barcode-printers/link-os/browser-print.html)


eLBati:
@sylvain-garancher @fmdl @JosDeGraeve what do you think?

gurneyalex (commented):
Can you provide information about the licence of the JS library that is shipped with this addon?

eLBati:
Hi @gurneyalex , thank you for pointing this out.
No license information is provided with the library, but I found this
https://www.zebra.com/us/en/support-downloads/support-eula/browser-print-eula-javascript.html

This is not compatible with OCA, right?

gurneyalex:
@eLBati I confirm this is not compatible (no right to modify the code of this file). I'm therefore closing this PR. Sorry about this.

PR #232, opened 2020-10-06
[IMP] base_report_to_printer: archive printers

jarroyomorales:
Allow to archive printers.

@etobella

### DIFF
--- a/base_report_to_printer/models/printing_printer.py
+++ b/base_report_to_printer/models/printing_printer.py
@@ -32,6 +32,7 @@ class PrintingPrinter(models.Model):
     _order = 'name'
 
     name = fields.Char(required=True, index=True)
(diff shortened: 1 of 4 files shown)


bodedra (approved):

etobella (approved):

etobella:
Can you prepare the cherry-pick for 13.0 please @jarroyomorales

pedrobaeza:
Shouldn't this be taken into account when updating printers from server (with `active_test=False` or similar for doing the match)?

bealdav (approved):

jarroyomorales:
@pedrobaeza After doing some testing I see it is necessary to add the `active_test=False` if we want to keep the archived printed updated. Thank you!

PR #233, opened 2020-10-08
[IMP] base_report_to_printer: archive printers

jarroyomorales:
Cherry-pick from #232 

@etobella @bodedra

### DIFF
--- a/base_report_to_printer/views/printing_printer.xml
+++ b/base_report_to_printer/views/printing_printer.xml
@@ -33,6 +33,13 @@
                     />
                 </header>
                 <sheet>
+                    <widget
+                        name="web_ribbon"
+                        text="Archived"
+                        bg_color="bg-danger"
+                        attrs="{'invisible': [('active', '=', True)]}"
+                    />
+                    <field name="active" invisible="1" />
                     <div class="oe_title">
                         <h1>
                             <field name="name" />
--- a/base_report_to_printer/models/printing_printer.py
+++ b/base_report_to_printer/models/printing_printer.py
@@ -31,6 +31,7 @@ class PrintingPrinter(models.Model):
     _order = "name"
 
     name = fields.Char(required=True, index=True)
+    active = fields.Boolean(default=True)
     server_id = fields.Many2one(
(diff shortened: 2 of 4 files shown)


base_report_to_printer/views/printing_printer.xml
```diff
@@ -33,6 +33,20 @@
...
+                    <div class="oe_button_box" name="button_box">
+                        <button
+                            name="toggle_active"
+                            type="object"
+                            class="oe_stat_button"
+                            icon="fa-archive"
+                        >
+                            <field
+                                name="active"
+                                widget="boolean_button"
```
etobella: Boolean button does not work on 13.0

etobella (changes requested):

base_report_to_printer/views/printing_printer.xml
jarroyomorales: True, fixed according to the new paradigm.

etobella (approved):

alexis-via (approved):

alexis-via:
Pedro's remark on the other PR seems important : 
<<
Shouldn't this be taken into account when updating printers from server (with active_test=False or similar for doing the match)?
>>

alexis-via:
It seems there is nothing to change, because in the method update_printers() on printing.server, the loop uses "server.printer_ids" (the O2M field), so it will also take into account the inactive printers.

jarroyomorales:
@alexis-via I tested the same on v13 and as I said in https://github.com/OCA/report-print-send/pull/232#issuecomment-745178190 we need to add the `active_test` context, since `server.printer_ids` doesnt return the archived printers.

Thank you for the review!

pedrobaeza:
Please fix pre-commit

jarroyomorales:
@pedrobaeza Its fixed here https://github.com/OCA/report-print-send/pull/240

alexis-via:
You're right server.printer_ids doesn't return the inactive printers (I was excepting the opposite !). I updated the v14 code too.

PR #235, opened 2020-10-27
[12.0][ADD] print_datamatrix

SimoRubi:
Add new module to print_datamatrix.

In Odoo core already exists type 'ECC200DataMatrix' that can be printed but it is very limited, also according to its own documentation in https://pypi.org/project/reportlab/ (installed with Odoo):
```
class ECC200DataMatrix(Barcode):
    '''This code only supports a Type 12 (44x44) C40 encoded data matrix.
    This is the size and encoding that Royal Mail wants on all mail from October 1st 2015.
    see https://bitbucket.org/rptlab/reportlab/issues/69/implementations-of-code-128-auto-and-data
    '''
```

Specifically, it limits the length of the value to be printed to 144 chars.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ addons:
       - python-yaml
       - cups
       - libcups2-dev
+      - libdmtx0a
 
 env:
   global:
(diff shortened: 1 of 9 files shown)


pedrobaeza:
Shouldn't this be `report_barcode_datamatrix`?

SimoRubi:
> Shouldn't this be `report_barcode_datamatrix`?

👍🏼 renamed

PR #240, opened 2020-12-15
[13.0][FIX] printer_zpl2: pylint disable fix

jarroyomorales:
The disable should be where it is executed.


etobella (approved):

etobella:
@pedrobaeza We might try to fast-track this. WDYT??

PR #239, opened 2020-12-15
[MIG][14.0] base_report_to_printer

alexis-via:
I included in this migration the active field on printing.printer as per PR https://github.com/OCA/report-print-send/pull/233

PR #242, opened 2021-01-11
[14.0] [MIG] base_report_to_printer: Migration to v14

CarlosRoca13:
cc @Tecnativa TT26425

This PR superseds https://github.com/OCA/report-print-send/pull/239 attending the comments and solving the travis problems

Please, @pedrobaeza @Yajo can you review this?


pedrobaeza:
Please comment when ready

CarlosRoca13:
@pedrobaeza I think it is ready yet

pedrobaeza:
I see several things:

- You can squash more the commit history, for example in a lot of Graeme's commits that are together.
- There's a merge commit in the commit history, which looks conflicting.
- There are some changes (like active field), that are included as part of the migration, and without respecting attribution by Alexis. You have to cherry-pick commit and remove it from the migration.
- Please squash as well migration commits by author.

Not sure if you have to redo the commit history from scratch.

CarlosRoca13:
@pedrobaeza can you review it again?

pedrobaeza (approved):
Fixing Alexis' email in latest commit

CarlosRoca13:
@pedrobaeza changed

base_report_to_printer/tests/test_printing_server.py
```diff
@@ -89,19 +89,6 @@ def test_update_printers_update_unavailable(self, cups):
...
-        rec_id = self.new_printer()
-        rec_id.toggle_active()
-        self.server.refresh()
-        cups.Connection().getPrinters().get.return_value = False
-        self.Model.action_update_printers()
-        self.assertEqual(
-            "unavailable",
-            rec_id.status,
-        )
-
```
yajo: Hummm this doesn't seem like a bad test that should be removed. Besides I see some changes in code related to this, so if you remove the test how do we know it is OK?

I mean: I don't understand why archived printers should still be searched, but if somebody took the time to even write a test for that, it seems like an important thing. 🤔 

Is there any reason to remove this?

base_report_to_printer/tests/test_report.py
```diff
@@ -2,14 +2,13 @@
...
-import mock
+from unittest import mock
 
 from odoo import exceptions
 from odoo.tests import common
 
 
-@common.at_install(False)
-@common.post_install(True)
+@common.tagged("post_install", "-at_install")
```
yajo: Unless strictly required, it is a good rule of thumb to avoid post-install tests, as they fail on integration environments more often than others.

Is it possible to remove this decorator?

base_report_to_printer/tests/test_res_users.py
```diff
@@ -4,8 +4,7 @@
 from odoo.tests import common
 
 
-@common.at_install(False)
-@common.post_install(True)
+@common.tagged("post_install", "-at_install")
```
yajo: Same about post-install

yajo (changes requested):
Some concerns...

CarlosRoca13:
@Yajo I applied the changes that you suggest. I recover the test that were deleted because the changes that @alexis-via made in this file https://github.com/OCA/report-print-send/pull/242/commits/89d3a6c5644bac8656a773ab1003445f29926e00#diff-c7ff1f639eeed9a13cafffb974ab10c275477ab876bab188862755ec706d6021

yajo (approved):

PR #244, opened 2021-01-20
[13.0] [FIX] base_report_to_printer: add dependency on mail

ThomasBinsfeld:
On existing installations, the cron creation fails because of the activity_user_type required field.
Adding a dependency on the mail module solved the problem.


Laurent-Corron (approved):

etobella (approved):

etobella:
@OCA/reporting-engine-maintainers I think this can be merged, WDYT?

PR #258, opened 2021-10-22
[FIX] base_report_to_printer: 1030: The printer or class does not exist

hugho-ad:
Before this commit a error 1030: The printer or class does not exist because cups tried to print using uri printer
now configuring cups.setServer and cups.setPort as recommended at
http://nagyak.eastron.hu/doc/system-config-printer-libs-1.2.4/pycups-1.9.51/html/cups.Connection-class.html

it's working well

### DIFF
--- a/base_report_to_printer/models/printing_server.py
+++ b/base_report_to_printer/models/printing_server.py
@@ -36,6 +36,8 @@ class PrintingServer(models.Model):
         self.ensure_one()
         connection = False
         try:
+            cups.setServer(self.address)
+            cups.setPort(self.port)
             connection = cups.Connection(host=self.address, port=self.port)
         except Exception:
             message = _(


hugho-ad:
@moylop260 @julioserna could you review pls?

luisg123v (approved):

moylop260:
It could have compatibility errors if an old version of pycups is installed

What about using `pycups>=1.9.51` from:
 - https://github.com/OCA/report-print-send/blob/29f6a27884e2a02ce284c366ea57479465292458/requirements.txt#L2

moylop260:
Could you check it, please?

@max3903 @gdgellatly
@CarlosRoca13 @igh-fp @schout-it

gdgellatly:
Hi @hugho-ad ,

This PR makes no sense to me in the context of the linked documentation.

```A connection to the CUPS server. Before it is created the connection server and username should be set using cups.setServer and cups.setUser; otherwise the defaults will be used```

So far so good. Now the very next line
```
The constructor takes optional arguments host, port, and encryption, which default to the values of cups.getServer(), cups.getPort(),
```

In this PR, you set the server and port, and then subsequently overwrite it in the very next argument where you construct the connection class. But they are set identically as before. So whatever is working or not, it cannot seem related to this or the linked documentation is wrong.

For me I need more specifics on why the existing code doesn't work and why this PR fixes it, as honestly, right now it just looks like 2 useless lines.

rousseldenis:
Done in https://github.com/OCA/report-print-send/pull/262

This is a duplicate

moylop260:
thank you @rousseldenis 

For record, issue related
 - https://github.com/OpenPrinting/pycups/issues/30

cc @gdgellatly

PR #259, opened 2021-11-09
[IMP] 12.0 base_report_to_printer: improve job title

gurneyalex:
instead of sending a temporary random file name as the title of the job,
use the report file name. Since this is not displayed by default by
cups, document how to configure the CUPS server so that the information
can be displayed.

Clean up the temporary file after printing to avoid cluttering the
server in the long run.

### DIFF
--- a/base_report_to_printer/models/ir_actions_report.py
+++ b/base_report_to_printer/models/ir_actions_report.py
@@ -4,8 +4,10 @@
 # Copyright (C) 2011 Domsense srl (<http://www.domsense.com>)
 # Copyright (C) 2013-2014 Camptocamp (<http://www.camptocamp.com>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import time
 
 from odoo import api, exceptions, fields, models, _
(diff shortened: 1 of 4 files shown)


base_report_to_printer/models/ir_actions_report.py
```diff
@@ -111,6 +113,14 @@ def print_document(self, record_ids, data=None):
             raise exceptions.Warning(
                 _('No printer configured to print this report.')
             )
+        report_file_names = [
+            safe_eval(self.print_report_name, {"object": obj, "time": time})
```
pedrobaeza: Is this `safe_eval` the same call as standard one with the same context? Maybe you can call to the same thing for not having possible differences?

gurneyalex: it is the same as the standard call, but the standard call is made in a controller (/report/download in addons/web) so it is not convenient to "call the same thing"

StephaneMangin (approved):

pedrobaeza (approved):

PR #261, opened 2021-11-09
[IMP] 14.0 base_report_to_printer: improve job title

gurneyalex:
instead of sending a temporary random file name as the title of the job,
use the report file name. Since this is not displayed by default by
cups, document how to configure the CUPS server so that the information
can be displayed.

Clean up the temporary file after printing to avoid cluttering the
server in the long run.


gurneyalex:
forward port of #259 #260

gurneyalex:
Merging as this is a forward port of #259 which is merged.

PR #264, opened 2021-11-24
[14.0][FWD] base_report_to_printer: Forward updates from 13.0

dzungtran89:
- ﻿﻿#244

### DIFF
--- a/base_report_to_printer/__manifest__.py
+++ b/base_report_to_printer/__manifest__.py
@@ -14,7 +14,7 @@
     " Open for Small Business Ltd",
     "website": "https://github.com/OCA/report-print-send",
     "license": "AGPL-3",
-    "depends": ["web"],
+    "depends": ["mail", "web"],
     "data": [
         "data/printing_data.xml",
         "security/security.xml",


pedrobaeza:
But I don't get the explanations of the reason of this dependency.

dzungtran89:
Hi @pedrobaeza , thanks for you feedback

I don't know the case that the author had dealt with, and not sure why that PR was merged too although there is module named [base_report_to_printer_mail](https://github.com/OCA/report-print-send/tree/13.0/base_report_to_printer_mail) which might fix the case.

So, in my opinion, we can close this porting PR and use the module `base_report_to_printer_mail` instead, its PR #254 has been created

nilshamerlinck:
The original commit in 13.0 states that:
> On existing installations, the cron creation fails because of the activity_user_type required field. Adding a dependency on the mail module solved the problem.

I could not reproduce the issue, could be a corner case ([default value on `activity_user_type`](https://github.com/odoo/odoo/blob/13.0/addons/mail/models/ir_actions.py#L41) not taken into account in a fresh registry with `mail` not loaded?). 

Note that `base_report_to_printer_mail` comes with the dependency to `mail` indeed, but was not aimed at fixing the issue above and comes with a change in the behavior in `base_report_to_printer` that might not suit everyone.

PR #270, opened 2022-01-01
[ADD] base_report_to_qz: new module to print with QZ Tray

angelmoya:
Module works, but could be nice to improve.

I push because I think that module could be good alternative to base report_to_printer, and maybe someone wants to colaborate.

QZ Tray allow to print over html, this way users can use local printers with Odoo.

You can install QZ Tray on Linux, Windows or Mac computer, and I'm not sure, but I read about use with android.

QZ Tray is LGPL 2.1 licensed, you can buy support to get certificates, but you can use your own certificates.

On Install.md I try to explain how to install, you need generate your own certificate, an configure in Odoo and QZ Tray.

Module works, but as I said it could be improve:
- Can print pdf or txt, I only tested txt as escpos raw printing... it's needed to test zpl or txt print.
- tests must be done
- pre-commit get errors, related with qz-tray.js that is the original library, and related with js code and requiremet OpenSSL.
- at this moment you need to write the name of the printer, could be nice to create a widget show local printers and select one (js code is quite simple).


chrisandrewmann:
@angelmoya Firstly i'd like to say a big thanks for your efforts on this module.
I was looking to create something similar and it would be a lot of help.
I'd like to chat about what we could do together on this module to get it to a state where it's usable for others and ideally OCA.
Feel free to reach out and maybe we can get somewhere.

I've migrated it to V15, V16 + V17 and it's working well, with some improvements:
- Made the user_id not required on printing.report.qz.tray - This way you can set a default for ALL users
- Updated to support new OWL action report handler and frontend logic
- Added ability to specify a remote server location (IP) for the printer, so it doesn't have to work only locally
- Tested with both PDF and ZPL using "Virtual ZPL Printer" - All looks good

angelmoya:
Hi @chrisandrewmann , can you do a pull request with version 16.0, maybe we can add and mantain this modules together

chrisandrewmann:
> Hi @chrisandrewmann , can you do a pull request with version 16.0, maybe we can add and mantain this modules together

Hi @angelmoya. Yes I would happily do this, which branch should I do the PR on? The main 16.0 branch only?

angelmoya:
16.0 would be great

chrisandrewmann:
PR has been created: https://github.com/OCA/report-print-send/pull/357
There are no unit tests though so it might struggle to get approved...
Let's hope it gains some traction because I really feel that an OCA QZ tray module could hugely benefit the community.

PR #271, opened 2022-01-05
[MIG] remote_report_to_printer: Migration to 14.0

olgamarcocb:
Standard Migration


remote_report_to_printer/models/printing_action.py
```diff
@@ -13,4 +13,6 @@ def _available_action_types(self):
         res.append(("remote_default", "Use remote's default"))
         return res
 
-    action_type = fields.Selection(selection=_available_action_types,)
+    action_type = fields.Selection(
```
etobella, line 16: I think this is not necessary (obviously the error comes from previous versions). With the extension of _available_acion_types it might be enough :smile:

olgamarcocb: This field is required. if it is removed, the data file cannot find it! 😉

etobella (approved):

andreampiovesana:
merge?

andreampiovesana (approved):
go

PR #275, opened 2022-03-04
[MIG] base_report_to_printer: Migration to 15.0

BT-dmontull:


gediminasvenc:
JS part is wrong, in v15 method _triggerDownload is working differently

bizzappdev (approved):
Code LGTM

LoisRForgeFlow:
> JS part is wrong, in v15 method _triggerDownload is working differently

@gediminasvenc Could you elaborate? What is the difference?

Thanks for reviewing.

gediminasvenc:
> > JS part is wrong, in v15 method _triggerDownload is working differently
> 
> @gediminasvenc Could you elaborate? What is the difference?
> 
> Thanks for reviewing.

Hi @LoisRForgeFlow, first there is an error in the browser's console:

image

This JS part is not super important i think, it is responsible for a pop-up warning, which informs about the printout result.

In v15 function _triggerDownload is no longer included in ActionManager, but now its included into another function called - makeActionManager. You can check it in the source code and compare.

ghost:
Bildschirmfoto 2022-06-14 um 10 34 59

I tried to find an usage snippet of "_triggerDownload". In the OCA module report_xlsx I found a comment that its not publically available on the service any more ...

thomas15v:
```javascript
/** @odoo-module */
import {registry} from "@web/core/registry";

async function cupsReportActionHandler(action, options, env) {
    if (action.report_type === "qweb-pdf") {
        const orm = env.services.orm;

        let print_action = await orm.call("ir.actions.report", "print_action_for_report_name", [action.report_name]);
        if (print_action && print_action.action === "server") {
            let result = await orm.call("ir.actions.report", "print_document",
                [action.id, action.context.active_ids, action.data]);
            if (result){
                env.services.notification.add(env._t("Successfully sent to printer!"));
            }
        }
        return true;
    }
}

registry
    .category("ir.actions.report handlers")
    .add("cups_report_action_handler", cupsReportActionHandler);
```

I fixed the javascript as followed. This javascript only works if you fix the safe_eval evaluation for document name in print_document, I couldn't be bothered and put everything about that name in commentary.

ghost:
So what else needs to be done to get it accepted?

Liquidarom:
Hi everybody,

i need ur help to use report_to_printer in my db. I use Odoo v15 on SH and i finished to configure an externaly server CUPS. 
When i try to update the printers, i have an error : Failed to connect to the CUPS server on **.***.***.***:631. Check that the CUPS server is running and that you can reach it from the Odoo server.
 
Do you know where the error comes from? please ;)

PR #283, opened 2022-06-08
[14.0][ADD] base_report_to_label_printer: Add default label printer

ne1800:
We have the case in the company that we have to print specific reports to a label printer. This should be done via the user printer. At the user it should be possible to set a label printer and a normal printer. At the report itself it should be defined if it is a label report.

This extension could perhaps help others.


base_report_to_printer/models/res_users.py
```diff
@@ -23,6 +23,9 @@ def _user_available_action_types(self):
     printing_printer_id = fields.Many2one(
         comodel_name="printing.printer", string="Default Printer"
     )
+    default_label_printer_id = fields.Many2one(
```
jbaudoux: That new field should also be added in `_register_hook` below.
We must allow the user to set a value in his preferences

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -59,9 +60,13 @@ def print_action_for_report_name(self, report_name):
     def _get_user_default_print_behaviour(self):
         printer_obj = self.env["printing.printer"]
         user = self.env.user
+        if self.label:
```
jbaudoux: move this in a function `_get_user_default_printer`

jbaudoux (changes requested):
It is better to add this in a new module `base_report_to_label_printer`
Just implement `_get_user_default_printer` in this module and extend it in the new module.

Can you also open a PR on https://github.com/OCA/wms/tree/14.0/shopfloor_workstation_label_printer
to depend on the new module and drop from `https://github.com/OCA/wms/blob/14.0/shopfloor_workstation_label_printer/models/res_users.py` what you moved here

jbaudoux:
cc @simahawk @sebalix

jbaudoux (approved):
LGTM

Can you squash the commits?

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -56,12 +56,16 @@ def print_action_for_report_name(self, report_name):
         }
         return serializable_result
 
+    def _get_user_default_printer(self, user):
+        return user.printing_printer_id
```
sebalix, line 60:
```suggestion
        self.ensure_one()
        return user.printing_printer_id
```
As we will check `self.label` in the override.

sebalix (approved):
One small remark, LG

jbaudoux:
cc @lmignon

TDu (approved):

simahawk:
@ne1800 can you rebase pls?

base_report_to_printer/models/ir_actions_report.py
simahawk: @ne1800 ping

mt-software-de:
> @ne1800 can you rebase pls?

@ne1800 is not available right now. 
I added his PR here https://github.com/jbaudoux/report-print-send/pull/2  maybe we can merge then both
@jbaudoux fyi

jbaudoux:
Included in https://github.com/OCA/report-print-send/pull/310
This PR can be closed

sebalix:
@mt-software-de @jbaudoux we should not mix these two PRs to separate concerns, I'm using this one in one project but not the other.
If @ne1800 is not available I can re-open this one elsewhere so it can be merged ASAP.

jbaudoux:
> @mt-software-de @jbaudoux we should not mix these two PRs to separate concerns, I'm using this one in one project but not the other. If @ne1800 is not available I can re-open this one elsewhere so it can be merged ASAP.

Let's approve and merge the other PR, then you can use the module you want without relying on open PR
I need to include this PR in the other one for the glue module

simahawk:
> > @mt-software-de @jbaudoux we should not mix these two PRs to separate concerns, I'm using this one in one project but not the other. If @ne1800 is not available I can re-open this one elsewhere so it can be merged ASAP.
> 
> Let's approve and merge the other PR, then you can use the module you want without relying on open PR I need to include this PR in the other one for the glue module

@jbaudoux hmm sounds like the contrary.... we must merge the base PR 1st, which is this one.
We can take this over and finish it.

simahawk:
there you go https://github.com/OCA/report-print-send/pull/313

PR #288, opened 2022-07-29
[IMP] report print send allow print multiple reports types

santostelmo:
Allow to print the same report in multiple report types  and behaviour (send to client or send to printer)

image

Currently we are only able to set 1 report type because the method `_get_report_from_name` it always takes the first report found by repor_name:
https://github.com/odoo/odoo/blob/0ab552a749e3fc4aa6c549909ccece3821c4cd09/odoo/addons/base/models/ir_actions_report.py#L498

The proposed solution is to send in the `.js` call the action linked to the report directly.

### DIFF
--- a/base_report_to_printer/models/ir_actions_report.py
+++ b/base_report_to_printer/models/ir_actions_report.py
@@ -41,12 +41,18 @@ class IrActionsReport(models.Model):
         self.printer_tray_id = False
 
     @api.model
-    def print_action_for_report_name(self, report_name):
+    def _get_report_from_id(self, report_id):
+        report_obj = self.env['ir.actions.report']
+        context = self.env['res.users'].context_get()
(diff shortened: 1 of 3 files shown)


base_report_to_printer/models/ir_actions_report.py
```diff
@@ -41,12 +41,19 @@ def onchange_printing_printer_id(self):
...
+        context = self.env['res.users'].context_get()
+        return report_obj.with_context(context).sudo().search(conditions, limit=1)
+
+    @api.model
+    def print_action_for_report_name(self, action_id):
         """Returns if the action is a direct print or pdf
 
         Called from js
         """
-        report = self._get_report_from_name(report_name)
```
santostelmo, line 49: This always get the fist `ir.actions.report`. It means that if we want to have multiple report types to be printed is not possible.

PR #292, opened 2022-08-25
[MIG] base_report_to_printer: Migration to 15.0

matiasperalta1:
superseed of https://github.com/OCA/report-print-send/pull/275


base_report_to_printer/__manifest__.py
```diff
@@ -0,0 +1,37 @@
...
+        "views/printing_server.xml",
+        "views/printing_job.xml",
+        "views/printing_report.xml",
+        "views/res_users.xml",
+        "views/ir_actions_report.xml",
+        "wizards/printing_printer_update_wizard_view.xml",
+    ],
+    "assets": {
+        "web.assets_backend": [
+            "/base_report_to_printer/static/src/js/qweb_action_manager.js",
```
suganthikarunanithi: The file name is wrong.
`base_report_to_printer/static/src/js/qweb_action_manager.esm.js`

suganthikarunanithi (changes requested):
Except the file name issue, the app looks fine.

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -110,7 +108,7 @@ def print_document(self, record_ids, data=None):
             raise exceptions.UserError(_("No printer configured to print this report."))
         if self.print_report_name:
             report_file_names = [
-                safe_eval(self.print_report_name, {"object": obj, "time": time})
```
etobella: Why do you remove the safe_eval?

etobella:
Seems to work the printing to cups, but it fails to show the PDF.

I made the following test on runboat:

- Installed account
- create a server and printer (I will not send it anyway)
- Modify invoice printing to use this server / printer
- Print an invoice --> It fails as  the server does not exist, but it tries it
- Print another report --> Nothing happens (no PDF or else)

Can you review the last point?

etobella:
@matiasperalta1 Do you need some help fixing JS?

base_report_to_printer/static/src/js/qweb_action_manager.esm.js
```diff
@@ -0,0 +1,31 @@
...
+                action.context.active_ids,
+                action.data,
+            ]);
+            if (result) {
+                env.services.notification.add(env._t("Successfully sent to printer!"));
+            } else {
+                env.services.notification.add(env._t("Could not sent to printer!"));
+            }
+        }
+        return true;
```
saxomanu: Don't return True on line 25 but between 23 and 24.
If a value is return in 25, all js stop, and report is not send to client if not server need

saxomanu (changes requested):

base_report_to_printer/models/ir_actions_report.py
matiasperalta1: I remove the safe_eval because the javascript does not work. Maybe it would have to be fixed for it to work well, but I didn't know how.

matiasperalta1:
@etobella I did the tests with another report and it worked.
I made the following test on runboat. 
1) create a server and printer
2) Modify Report Layout Preview printing to use this server / printer
3) Print a Report Layout Preview --> It fails as the server does not exist, but it tries it

etobella:
Try it without setting the server. Nothing happens. no pdf is generated. Tested on runbot

etobella (approved):
Great!!! Now it works as expected!!!
LGTM
:+1:

base_report_to_printer/models/ir_actions_report.py
etobella: Checking Odoo code, this is necessary. You can review it here:

https://github.com/odoo/odoo/blob/c9980b8f946005904cf4701b71a1c37e7577ea65/addons/web/controllers/main.py#L2032-L2034

matiasperalta1:
@etobella Thanks for you review!
Who could be able to mark the addon to be migrated?

etobella:
There is still the safe_eval change to make it similar to odoo standard. Can you review that?

matiasperalta1:
@etobella safe_eval fixed

etobella (approved):
Just one last thing and please, change the commit name for something like:

`[MIG] base_report_to_printer: Finish migration to 15.0`

Thanks!!!

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -0,0 +1,161 @@
...
+            must_skip_send_to_printer=True
+        )._render_qweb_pdf(record_ids, data=data)
+        behaviour = self.behaviour()
+        printer = behaviour.pop("printer", None)
+
+        if not printer:
+            raise exceptions.UserError(_("No printer configured to print this report."))
+        if self.print_report_name:
+            report_file_names = [
+                safe_eval(self.print_report_name, {"object": obj})
```
etobella: If you import time from odoo.tools.safe_eval you can keep the time :wink: 
[from odoo.tools.safe_eval import safe_eval, time](https://github.com/odoo/odoo/blob/c9980b8f946005904cf4701b71a1c37e7577ea65/addons/web/controllers/main.py#L40)

igh-odoo:
Hi,

great work! We are really looking forward to the 15.0 migration.

I installed the current version from the PR today and what I noticed was that users that do not have the Employees/Officer role are not allowed to edit their own profile any more:

The requested operation can not be completed due to security restrictions.

Document type: Users (res.users)
Operation: read
User: 14
Fields:

    address_home_id (allowed for groups 'Employees / Officer')
    barcode (allowed for groups 'Employees / Officer')
    birthday (allowed for groups 'Employees / Officer')
    certificate (allowed for groups 'Employees / Officer')
    children (allowed for groups 'Employees / Officer')
    country_of_birth (allowed for groups 'Employees / Officer')
    emergency_contact (allowed for groups 'Employees / Officer')
    emergency_phone (allowed for groups 'Employees / Officer')
    employee_bank_account_id (allowed for groups 'Employees / Officer')
    employee_country_id (allowed for groups 'Employees / Officer')
    employee_phone (allowed for groups 'Employees / Officer')
    gender (allowed for groups 'Employees / Officer')
    identification_id (allowed for groups 'Employees / Officer')
    km_home_work (allowed for groups 'Employees / Officer')
    marital (allowed for groups 'Employees / Officer')
    passport_id (allowed for groups 'Employees / Officer')
    permit_no (allowed for groups 'Employees / Officer')
    pin (allowed for groups 'Employees / Officer')
    place_of_birth (allowed for groups 'Employees / Officer')
    spouse_birthdate (allowed for groups 'Employees / Officer')
    spouse_complete_name (allowed for groups 'Employees / Officer')
    study_field (allowed for groups 'Employees / Officer')
    study_school (allowed for groups 'Employees / Officer')
    visa_expire (allowed for groups 'Employees / Officer')
    visa_no (allowed for groups 'Employees / Officer')

I think this has to to do something with the module adding fields to the res.users model, hasn't it?

Best regards,
Florian Pose

igh-odoo:
> I think this has to to do something with the module adding fields to the res.users model, hasn't it?

No. I noticed, that the error is gone if I disable the  `res.users.form.simple (in base_report_to_printer)` view.

base_report_to_printer/models/ir_actions_report.py
matiasperalta1: Thanks you very much @etobella for your review

etobella:
@igh-odoo You are right, tested on runboat. It fails when we install hr and base_report_to_printer.

etobella:
Probably related to this:

https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/models/res_users.py#L264-L274

Do something similar to https://github.com/odoo/odoo/blob/15.0/addons/google_calendar/models/res_users.py#L30-L36 and all should work properly

matiasperalta1:
@pedrobaeza Hi Pedro. It would seem that everything works fine. Do you want to check it and if everything works fine merge it?

pedrobaeza:
What about this comment?

https://github.com/OCA/report-print-send/pull/292#issuecomment-1241610027

matiasperalta1:
@pedrobaeza I think it is not related to a problem of this module.

pedrobaeza:
@etobella do you agree?

igh-odoo:
> @pedrobaeza I think it is not related to a problem of this module.

The error ist still present in 0dee1ac9e5e5ed32cb948ce95498e0f1fb0c7b1d.

If `res.users.form.simple` is deactivated, everything is fine.

base_report_to_printer/models/res_users.py
```diff
@@ -0,0 +1,42 @@
...
+            if code != "user_default"
+        ]
+
+    printing_action = fields.Selection(selection=_user_available_action_types)
+    printing_printer_id = fields.Many2one(
+        comodel_name="printing.printer", string="Default Printer"
+    )
+
+    @api.model
+    def _register_hook(self):
```
etobella: This hook is no longer working, do the change we commented please. You can test it installing hr and this module on the same database.

etobella (changes requested):

base_report_to_printer/models/res_users.py
matiasperalta1: Everything seems work fine! @etobella @pedrobaeza

pedrobaeza: OK, I need the others to confirm this

igh-odoo: Just tested. It works. Thank you!

pedrobaeza: @etobella ok by your part?

matiasperalta1:
@etobella Hi! about this method everything seems work fine. Do you agree?

igh-odoo:
Any idea, when this can be merged? Is it blocked because people did not sign the CLA, or other reasons?

base_report_to_printer/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+from . import ir_actions_report
+from . import printing_action
+from . import printing_job
+from . import printing_printer
+from . import printing_server
```
duyanhk15, line 5: I think, it should be
```suggestion
from . import printing_server
from . import printing_printer
from . import printing_job
```

base_report_to_printer/models/__init__.py
```diff
@@ -0,0 +1,8 @@
...
+from . import printing_report_xml_action
+from . import printing_tray
+from . import res_users
```
duyanhk15, line 8: It should:
```suggestion
from . import ir_actions_report
from . import printing_report_xml_action
from . import printing_tray
from . import res_users
```

yajo:
> Any idea, when this can be merged? Is it blocked because people did not sign the CLA, or other reasons?

@etobella should respond in https://github.com/OCA/report-print-send/pull/292#discussion_r975016827 if I'm not wrong.

etobella (approved):
Thanks!! Now it works as expected

PR #294, opened 2022-09-02
[ADD] printer_escpos

etobella:
This modules allows to generate escpos tickets using XML

The idea was taken from https://github.com/fvdsn/py-xml-escpos but
it was rewritten in order to make it easier to manage (Licensed with MIT)

Usage
---------

To print a ticket, you need to call the ticket printing method.
It can be used anywhere (other modules, server actions, etc.).

Example : Print a ticket of a product ::

    self.env['printing.escpos'].browse(ticket_id).print_escpos(
        self.env['printing.printer'].browse(printer_id),
        self.env['product.product'].browse(product_id))

You can also use the generic ticket printing wizard, if added on some models.

Format of the ticket
--------

The ESCPOS expects an XML with root tag receipt.
Inside we can use the following tags:

*  p
*  div
*  section
*  article
*  receipt
*  header
*  footer
*  li
*  h1
*  h2
*  h3
*  h4
*  h5
*  line: Allows to manage lines. Each column must be right or left
*  ul: Bullet list
*  ol: Ordered list
*  pre
*  hr
*  br: Line break
*  qr: Inserts a QR. The text will be the value of the QR. With attribute size we can make it bigger
*  img: Inserts an image
*  barcode: Inserts a barcode. Expects an attribute encoding in order to know which kind of barcode
*  cut
*  partialcut
*  cashdraw: Opens the cashdraw if configured

At any tag we can add the following attributes in order to modify the style:

*  align
*  underline: Can be on or off
*  bold: Can be on or off
*  size: Can be normal, double, double-height or double-weight
*  font: Can be A or B
*  width
*  indent
*  tabwidth
*  bullet: Describes the bullet description. By default ` - `
*  line-ratio
*  color


sanchonuria:
We have tested it and it works mainly, but I found something that fails. I don't know if I am doing all properly. I used the following x ml:

```xml
<receipt>
    <h1>Receipt!</h1>
    <h2>div,span,p,ul,ol are also supported</h2>:w
    <line>
        <left>Product</left>
        <right>0.15€</right>
    </line>
    <hr />
    <line size='double-height'>
        <left>TOTAL</left>
        <right>0.15€</right>
    </line>
    <barcode encoding='ean13'>
        5449000000996
    </barcode>
    <cashdraw /> 
    <cut />
</receipt>
```

With the following result:
ticket


Also, a demo would be great :wink:

etobella:
Sure!! I added some text, I hope now it works properly :smile: 
lines only work with 2 columns right now.

printer_escpos/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "author": "CreuBlanca,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/report-print-send",
+    "depends": ["base_report_to_printer"],
+    "data": [
+        "wizards/print_record_escpos.xml",
+        "security/ir.model.access.csv",
+        "views/printing_escpos.xml",
+    ],
+    "demo": ["demo/demo.xml"],
+    "external_dependencies": {"python": ["python-escpos"]},
```
sanchonuria, line 18: Shouldn't be the library `escpos` with your last changes?

etobella: Nope, the right one is to use this one, but we need to get version 3 or more. However, they are on pre-release

https://pypi.org/project/python-escpos/#history

etobella:
Error on ZPL2 is fixed on #296

sanchonuria (approved):

PR #290, opened 2022-08-02
[10.0][IMP] pingen api v2

grindtildeath:

PR #303, opened 2022-10-21
[10.0][IMP] pingen api v2

ajaniszewska-dev:
supersedes: https://github.com/OCA/report-print-send/pull/290/files


ajaniszewska-dev:
travis fails on lint dependencies

PR #304, opened 2022-11-02
[15.0][MIG] printer_zpl2: Migration to 15.0

duyanhk15:


printer_zpl2/models/printing_label_zpl2.py
```diff
@@ -123,10 +123,8 @@ def check_recursion(self):
 
     def _get_component_data(self, record, component, eval_args):
         if component.data_autofill:
-            data = component.autofill_data(record, eval_args)
-        else:
-            data = component.data
-        return safe_eval(str(data), eval_args) or ""
+            return component.autofill_data(record, eval_args)
```
etobella, line 126: Why this change?

etobella (commented):
Only a small question, all the other seems fine IMO :+1:

printer_zpl2/models/printing_label_zpl2.py
hailangvn: Good day @etobella, thanks for your comment. We faced below error because `str(data)` convert `datetime` value to `datetime.datetime(2022, 10, 28, 8, 58, 26, 226475)` format. So instead of doing unnecessary work (convert to string then convert back to python), we just simply return the data. The `safe_eval` is still useful in the line after above line to convert string from db.
```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/opt/OCA/report-print-send/tree/15.0/printer_zpl2/models/printing_label_zpl2.py", line 129, in _get_component_data
    return safe_eval(str(data), eval_args) or ""
  File "/opt/odoo/15.0/odoo/tools/safe_eval.py", line 348, in safe_eval
    raise ValueError('%s: "%s" while evaluating\n%r' % (ustr(type(e)), ustr(e), expr))
ValueError: <class 'NameError'>: "name 'datetime' is not defined" while evaluating
"{'id': 185, 'create_date': datetime.datetime(2022, 10, 28, 8, 58, 26, 226475)}"
```

cyrilmanuel (approved):

yankinmax (approved):

MiquelRForgeFlow:
@duyanhk15 I think you should cherry-pick https://github.com/OCA/report-print-send/pull/308 to solve the merging issue. And also https://github.com/OCA/report-print-send/pull/319 to follow previous versions.

pedrobaeza:
```
printer_zpl2/wizard/wizard_import_zpl2.py:58: [W0109(duplicate-key), _default_font_format] Duplicate key 'zpl2.ARG_HEIGHT' in dictionary
```

MiquelRForgeFlow:
As I said before, https://github.com/OCA/report-print-send/pull/308 has to be cherry-picked.

pedrobaeza:
Your comment is misleading, as you are pointing to the pre-commit update PR. Point instead to the specific commit doing that.

MiquelRForgeFlow:
My comment is not misleading. I suppose that whoever sees that PR will see two commits, and will deduce that one is a pre-commit update (which is not needed) and the other one is the needed fix to be cherry-picked. Yeah, I could put the hash here of that commit, but that's two more clicks that putting the link of the PR.

pedrobaeza:
This link makes both of them in one shot and less misleading: https://github.com/OCA/report-print-send/pull/308/commits/094c4ec6d2b2f5401490aec78c2094691b8ff644

nilshamerlinck:
Thanks guys, cherry-picked missing commits here: https://github.com/OCA/report-print-send/pull/320

PR #305, opened 2022-11-02
15.0 mig remote report to printer

ghost:
upgrade with minimal changes


ghost:
Hi, I get the following error:
ImportError: cannot import name 'formatargspec' from 'inspect' 
is it possible that this error is not connected to the module?

PR #315, opened 2023-01-26
[16.0][MIG] pingen: Migration to 16.0

ajaniszewska-dev:
includes changes from: https://github.com/OCA/report-print-send/pull/290/files

image
image

PR #327, opened 2023-04-28
[ADD] pingen_env

ajaniszewska-dev:
we want to manage pingen config values via server environment


pingen_env/models/res_company.py
```diff
@@ -0,0 +1,45 @@
...
+    @api.depends()
+    def _compute_pingen_env(self):
+        global_section_name = "pingen"
+        for company in self:
+            # default vals
+            config_vals = {
+                "pingen_clientid": "",
+                "pingen_client_secretid": "",
+                "pingen_organization": "",
+                "pingen_staging":False,
```
grindtildeath: Actually, I think we want a `True` value as default here, to avoid sending anything to prod without setting it explicitly.

pingen_env/views/base_config_views.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+   <record model="ir.ui.view" id="res_config_settings_view_form">
+    <field name="model">res.config.settings</field>
+    <field name="inherit_id" ref="pingen.res_config_settings_view_form" />
+    <field name="arch" type="xml">
+      <field name="pingen_clientid" position="attributes">
+        <attribute name="attrs" eval="False" />
```
grindtildeath: Are these overrides needed? I don't see any `attrs` defined in https://github.com/OCA/report-print-send/pull/315/files#diff-cdd40590b017ab65f0a06265805b28d4a3cf08b24966d0f5211a6167441758f6R17

pingen_env/hooks.py
```diff
@@ -0,0 +1,15 @@
...
+
+def post_init_hook(cr, registry):
+    # When installing pingen_env we want to delete data from DB
+    cr.execute(
+        """UPDATE res_company
+        SET pingen_clientid = NULL,
+        pingen_client_secretid = NULL,
+        pingen_organization = NULL,
+        pingen_staging = NULL,
+        pingen_webhook_secret = NULL
```
grindtildeath: Nitpicking
```suggestion
        pingen_webhook_secret = NULL;
```

grindtildeath (approved):

pingen_env/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+{
+    "name": "pingen.com server environment",
+    "version": "16.0.1.0.0",
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "Reporting",
+    "complexity": "easy",
+    "depends": ["pingen", "server_environment"],
+    "website": "https://github.com/OCA/report-print-send",
+    "data": [],
```
TDu: empty keys can be removed

TDu (approved):

grindtildeath:
@ajaniszewska-dev can you please fix the tests? :pray:

ajaniszewska-dev:
@pedrobaeza  can we merge? :pray:

pingen_env/README.rst
```diff
@@ -0,0 +1 @@
+to update
```
pedrobaeza: This is not correct

pingen_env/hooks.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+
+def post_init_hook(cr, registry):
+    # When installing pingen_env we want to delete data from DB
+    cr.execute(
+        """UPDATE res_company
```
pedrobaeza: This seems a solution of your current problem in existing DBs instead of a generic one.

ajaniszewska-dev: @pedrobaeza  updated readme + removed hook, i think we should be ok now

pingen_env/readme/USAGE.rst
```diff
@@ -0,0 +1,23 @@
+Define values under global section name called 'pingen', or in case of different config per company - global section name connected with company name:  
+
+
+[pingen.companyA]
```
pedrobaeza: Isn't there any problem with spaces in the company name or any other special character not correctly being parsed by configuration helper?

And in any case, it should be more descriptive if you put a real demo company name:

```suggestion
[pingen.My Company (San Francisco)]
```

ajaniszewska-dev: agree company name maybe not the best way to differentiate, but we use it in prod env for many months already so for now ill leave it as it is.
Updated readme.

simahawk: but... please add a todo to use a tech name. This can be fragile.

pingen_env/readme/USAGE.rst
```diff
@@ -0,0 +1,23 @@
+Define values under global section name called 'pingen', or in case of different config per company - global section name connected with company name:  
+
+
+[pingen]
```
simahawk: this must be wrapped in code block

ajaniszewska-dev:
@pedrobaeza  good to merge?

pedrobaeza:
@simahawk must say, as he has comments.

simahawk (approved):

PR #346, opened 2024-01-31
[15.0][IMP] base_report_to_printer: out of connection fallback to client

chienandalu:
If the CUPS server isn't available the user won't be able to do anything to print the report they need.

At last we can give them the chance to have a fallback behavior downloading the document.

cc @Tecnativa TT47134

take a look @sergio-teruel


sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #360, opened 2024-07-25
[FIX] base_report_to_printer:  add missing context to orm call

BT-dmoreno:
Context should be passed to orm.call. This might be important if the target report is using a special report class to customize content of xml report based on context variables.


chienandalu (commented):
Can you rebase and push again?

fyi @sergio-teruel

PR #367, opened 2024-10-09
[16.0][IMP] base_report_to_printer: out of connection fallback to client

chienandalu:
fw of #346 

If the CUPS server isn't available the user won't be able to do anything to print the report they need.

At last we can give them the chance to have a fallback behavior downloading the document.

cc @Tecnativa TT47134

please review @sergio-teruel @carlosdauden


chienandalu:

carlosdauden (approved):

sergio-teruel (approved):

PR #370, opened 2024-10-14
[17.0][FW] oca-port: forward missing commits from 14.0

trisdoan:
### This change
- Took this occasion to 
  - fix access right: 
     - Access rights were added in https://github.com/OCA/report-print-send/commit/c7fac1bd4346828bb41012eab00caf86a392c421
     - But the author missed to add create access for the wizard, which results in Access Error when clicking 'Update Printers from CUPS'
  - fix wrong translated term

### DIFF
--- a/.oca/oca-port/blacklist/base_report_to_printer.json
+++ b/.oca/oca-port/blacklist/base_report_to_printer.json
@@ -1,0 +1,5 @@
+{
+  "pull_requests": {
+    "242": "(auto) Nothing to port from PR #242"
+  }
+}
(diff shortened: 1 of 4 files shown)


sebalix:
Commit 04695e7fe356b41a4ab5441e360167b0685cc0be should not be ported, it is adding a whole module so it's not a changeset to port but to will have to be migrated in its own PR later.

sebalix (changes requested):
Commit 04695e7fe356b41a4ab5441e360167b0685cc0be should not be ported, it is adding a whole module so it's not a changeset to port but to will have to be migrated in its own PR later.

PR #371, opened 2024-10-14
[18.0][MIG] base_report_to_printer: Migration to 18.0

trisdoan:
### Note
- Missing commits from 14.0 are ported in https://github.com/OCA/report-print-send/pull/370
- Tests are failed due to lack of CUPS server configured

### Changes in 18.0
- `ir.property` was removed in https://github.com/odoo/odoo/commit/de302c2d36305c0d7562572a30587641eabfe914. Hence, replace default value for `property_printing_action_id` with invoked function.
- `doall` & `numbercall` fields were removed from ir_cron in https://github.com/odoo/odoo/pull/151628/commits/3a9949adbc87a1ca835aa7688e1f426b0ebbc8a8
- callable is not validated in selection any longer due to performance cost in https://github.com/odoo/odoo/pull/154177/commits/eaef3d28236fed7c72da287f060ae37aa8617f01. So, initialise `AVAILABLE_ACTION_TYPES` with the precomputed value instead of the function which returns the precomputed value.

### This changes
- Take this chance to replace percent format with f-string


sebalix:
Have you a clue why CUPS is not running in the tests? (I compared the `test.yml` GH workflow file with the 17.0, and didn't saw real difference...)

nilshamerlinck:
- these tests don't need a real cups server to run, as calls are mocked
- what is new is that 18.0 repos are now configured with `OCA_ENABLE_CHECKLOG_ODOO=1`, so the warnings are blocking
- but these `WARNING odoo odoo.addons.base_report_to_printer.models.printing_server: Failed to connect to the CUPS server on localhost:631. Check that the CUPS server is running and that you can reach it from the Odoo server. ` are actually expected, as concerned tests are testing some failure cases
- so @trisdoan could you please use the `.assertLogs` context manager to trap these logs? see [here](https://github.com/OCA/server-tools/commit/7d2e3d88c20175a91fcd5cf223417d9cf5da2f49) for an example

base_report_to_printer/models/printing_action.py
```diff
@@ -5,22 +5,22 @@
...
     _description = "Print Job Action"
 
-    @api.model
-    def _available_action_types(self):
-        return [
-            ("server", "Send to Printer"),
-            ("client", "Send to Client"),
-            ("user_default", "Use user's defaults"),
-        ]
+    _available_action_types = AVAILABLE_ACTION_TYPES
```
ivantodorovich: I think I liked the previous code better (here, and in all other similar cases), as it allowed for easier extension through overrides.
I wonder, what motivated this change?

ivantodorovich (approved):
non blocking comment
pre-approving: just the tests to fix

trisdoan:
~~DRAFT: update test~~

base_report_to_printer/models/printing_action.py
trisdoan: Hi @ivantodorovich, thanks for your in-depth review

As Odoo native decided to bypass the validation when the selection is a callable in [here](https://github.com/odoo/odoo/pull/154177/commits/eaef3d28236fed7c72da287f060ae37aa8617f01), it breaks some logic of the module, e.g: 
https://github.com/OCA/report-print-send/blob/58cba795f0555c64798507cf3298ac6b05d2f8ee/base_report_to_printer/tests/test_res_users.py#L15-L19

But my approach supports extension as well, no?

sebalix: Could it work with a property? So we keep a method to ease the override

sebalix (approved):
LG

chienandalu:
I forwarded several commits up to 17.0, could you update your migration base?

trisdoan:
> I forwarded several commits up to 17.0, could you update your migration base?

Hi @chienandalu, it's done

base_report_to_printer/models/printing_action.py
trisdoan: Hi @sebalix, using 'property' is a good approach  :+1: , thank you

I updated the code

base_report_to_printer/static/src/js/qweb_action_manager.esm.js
```diff
@@ -0,0 +1,85 @@
...
+            env.services.notification.add(_t("Could not sent to printer!"), {
+                type: "danger",
+            });
+            printer_exception = true;
+        }
+        if (print_action && print_action.action === "server" && printer_exception) {
+            // Just so the translation engine detects them as it doesn't do it inside
+            // template strings
+            const terms = {
+                the_report: env._t("The report"),
```
chienandalu:
```suggestion
                the_report: _t("The report"),
```

base_report_to_printer/static/src/js/qweb_action_manager.esm.js
```diff
@@ -0,0 +1,85 @@
...
+                couldnt_be_printed: env._t(
```
chienandalu:
```suggestion
                couldnt_be_printed: _t(
```

base_report_to_printer/static/src/js/qweb_action_manager.esm.js
```diff
@@ -0,0 +1,85 @@
...
+                    "couldn't be printed. Click on the button below to download it"
+                ),
+                issue_on: env._t("Issue on"),
```
chienandalu:
```suggestion
                issue_on: _t("Issue on"),
```

base_report_to_printer/static/src/js/qweb_action_manager.esm.js
```diff
@@ -0,0 +1,85 @@
...
+                    `<p>${terms.the_report} <strong>${action.name}</strong> ${terms.couldnt_be_printed}</p>`
+                ),
+                {
+                    title: `${terms.issue_on} ${print_action.printer_name}`,
+                    type: "warning",
+                    sticky: true,
+                    messageIsHtml: true,
+                    buttons: [
+                        {
+                            name: env._t("Print"),
```
chienandalu:
```suggestion
                            name: _t("Print"),
```

chienandalu (changes requested):

sebalix:
Hello @chienandalu , are you OK with the latest changes?

james-willdooit (approved):

PR #373, opened 2024-10-18
[15.0][FIX] base_report_to_printer: Report send to printer when is rendered from email template

sergio-teruel:
cc @Tecnativa TT51079

ping @chienandalu @pilarvargas-tecnativa

### DIFF
--- a/base_report_to_printer/models/mail_template.py
+++ b/base_report_to_printer/models/mail_template.py
@@ -1,0 +1,19 @@
+# Copyright 2024 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class MailTemplate(models.Model):
+    _inherit = "mail.template"
+
+    def generate_email(self, res_ids, fields):
+        """Ensure that any report rendered to send it by email not is sent to printer
+        directly if the report is setting to send it to printer.
+        For example, user set delivery slip report redirect to printer and
+        'Send an email when picking is validated' is active, if we do not inject this
+        context the report will be send twice to printer.
+        """
+        return super(
+            MailTemplate, self.with_context(must_skip_send_to_printer=True)
+        ).generate_email(res_ids, fields)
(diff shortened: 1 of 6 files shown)


base_report_to_printer/models/mail_template.py
```diff
@@ -0,0 +1,19 @@
...
+    def generate_email(self, res_ids, fields):
+        """Ensure that any report rendered to send it by email not is sent to printer
+        directly if the report is setting to send it to printer.
+        For example, user set delivery slip report redirect to printer and
+        'Send an email when picking is validated' is active, if we do not inject this
+        context the report will be send twice to printer.
+        """
+        return super(
+            MailTemplate, self.with_context(must_skip_send_to_printer=True)
+        ).generate_email(res_ids, fields)
```
pilarvargas-tecnativa, line 19: I believe this solution is already covered in a module that is not migrated.
https://github.com/OCA/report-print-send/blob/14.0/base_report_to_printer_mail/models/mail_template.py

I don't know if it is convenient to keep it separate and migrate the other module.

sergio-teruel:
What?? There is a module to fix other module??
I think that if you install base_report_to_printer in any instance the mail module allways will be installed, this is the reason by adding web dependency and fix it inside the module...

pilarvargas-tecnativa:
> What?? There is a module to fix other module?? I think that if you install base_report_to_printer in any instance the mail module allways will be installed, this is the reason by adding web dependency and fix it inside the module...

Then that module should disappear with a merge script. I leave it as a comment in case someone uses it, not in our case.

pilarvargas-tecnativa (approved):

pedrobaeza (changes requested):
That module exists because the dependency in this module is `web`, not `mail`. You may say that both modules are always installed, but that's not really something that you can assure, so I think that changing the dependency mustn't be done here, but on a new version, and for now, to migrate the other module.

chienandalu (commented):
I agree that is quite absurd to have that glue module, but let's migrate this and move on...

sergio-teruel:
Migration of module https://github.com/OCA/report-print-send/pull/374

PR #375, opened 2024-10-26
[15.0][IMP] base_report_to_printer: Add printer option to launch reports in new thread

sergio-teruel:
cc @Tecnativa TT51078

ping @chienandalu @CarlosRoca13 @pilarvargas-tecnativa

### DIFF
--- a/base_report_to_printer/models/ir_actions_report.py
+++ b/base_report_to_printer/models/ir_actions_report.py
@@ -3,9 +3,11 @@
 # Copyright (C) 2011 Agile Business Group sagl (<http://www.agilebg.com>)
 # Copyright (C) 2011 Domsense srl (<http://www.domsense.com>)
 # Copyright (C) 2013-2014 Camptocamp (<http://www.camptocamp.com>)
+# Copyright 2024 Tecnativa - Sergio Teruel
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import threading
 
-from odoo import _, api, exceptions, fields, models
+from odoo import _, api, exceptions, fields, models, registry
 from odoo.tools.safe_eval import safe_eval, time
 
 REPORT_TYPES = {"qweb-pdf": "pdf", "qweb-text": "text"}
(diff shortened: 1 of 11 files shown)


base_report_to_printer/models/printing_printer.py
```diff
@@ -66,6 +66,7 @@ class PrintingPrinter(models.Model):
     tray_ids = fields.One2many(
         comodel_name="printing.tray", inverse_name="printer_id", string="Paper Sources"
     )
+    multi_thread = fields.Boolean()
```
chienandalu: Why not do it at server level?

sergio-teruel: Changed to server level

pilarvargas-tecnativa:
Hi! @sbidoul this runbot is failed too

sbidoul:
@pilarvargas-tecnativa runboat looks ok now.

pedrobaeza:
Confirmed that is working again in other PR.

pilarvargas-tecnativa (approved):

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge minor

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -118,6 +120,29 @@ def behaviour(self):
...
         return result
 
+    def print_document_client_action(self, record_ids, data=None):
+        behaviour = self.behaviour()
+        printer = behaviour.pop("printer", None)
+        if printer.multi_thread:
+
+            @self.env.cr.postcommit.add
+            def _launch_print_thread():
+                threaded_calculation = threading.Thread(
```
lmignon, line 130: @sergio-teruel Why do you need to create a new thread? It could lead to the creation of ghosts threads since they are not supervised....

lmignon: I think I understand the use case. In fact, by launching post-commit printing, you solve another problem, namely that there should be no printing if the current transaction fails. Reading the code, I imagined that it was this last problem that you were trying to solve. Hence my question about using another thread for printing and making this fix conditional on authorizing printing in another thread.

sergio-teruel: Hi @lmignon Thanks for review.

I will try to explain the use case...

You Imagine that you are in a warehouse reading with a barcode devide de incoming goods... well, I have some modules that send labels to cups server automatically, if I do not send the report to cups in other thread the user has to wait for the label to be rendered, after that the label it's sent to cups... this operation takes about 1.5 seconds at best.

This is the reason to use (If you want) the print in other thread... I know that the time is the same, but the user has no wait each rendering operation....

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -118,6 +120,29 @@ def behaviour(self):
...
+                )
+                threaded_calculation.start()
+
+            return True
+        else:
+            return self.print_document(record_ids, data=data)
+
+    def print_document_threaded(self, report_id, record_ids, data):
+        with registry(self._cr.dbname).cursor() as cr:
+            self = self.with_env(self.env(cr=cr))
```
lmignon, line 142: you should not reuse the env into another thread.....

sergio-teruel: Oks.. 

So, I can create a new one...

base_report_to_printer/models/ir_actions_report.py
lmignon: IMO In this case you should use queue_job but not spawn another process :thinking:

PR #378, opened 2024-11-11
[15.0][IMP] base_report_to_printer: exceptions notifications

chienandalu:
Better handling of exceptions feedback. A notification will show up with the issued printer and report and a button for the user to download the report as a fallback to the failure.

Peek 11-11-2024 17-31

cc @Tecnativa TT51628

please review @sergio-teruel @carlosdauden

### DIFF
--- a/base_report_to_printer/i18n/es.po
+++ b/base_report_to_printer/i18n/es.po
@@ -1,23 +1,21 @@
 # Translation of Odoo Server.
 # This file contains the translation of the following modules:
-# * base_report_to_printer
+# 	* base_report_to_printer
 #
-# Translators:
-# OCA Transbot <transbot@odoo-community.org>, 2017
 msgid ""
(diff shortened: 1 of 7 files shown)


base_report_to_printer/i18n/es.po
```diff
@@ -657,7 +666,7 @@ msgstr "Lista de impresoras"
 #: model:ir.ui.menu,name:base_report_to_printer.printing_menu
 #: model_terms:ir.ui.view,arch_db:base_report_to_printer.view_users_form
 msgid "Printing"
-msgstr "Impresión"
+msgstr "Imprimiendo"
```
chienandalu, line 669: Looks like my system is introducing bad translations... :/

carlosdauden (approved):

sergio-teruel (approved):

PR #389, opened 2025-02-03
[IMP][16.0] remote_report_to_printer and base_report_to_printer: fix colsapn and keyerro translation

kluna1998:
fix colspan (remote_report_to_printer)
fix keyerror in .po (base_report_to_printer)

@etobella

### DIFF
--- a/base_report_to_printer/i18n/tr.po
+++ b/base_report_to_printer/i18n/tr.po
@@ -985,4 +985,4 @@ msgstr "yazdırılamadı. İndirmek için aşağıdaki butona tıklayın"
 #: code:addons/base_report_to_printer/wizards/print_attachment_report.py:0
 #, python-format
 msgid "{name} ({copies} copies)"
-msgstr "{isim} ({kopyalar} kopyalar)"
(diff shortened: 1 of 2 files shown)


remote_report_to_printer/views/res_remote_views.xml
```diff
@@ -7,7 +7,7 @@
         <field name="arch" type="xml">
             <notebook position="inside">
                 <page name="printers" string="Printers">
-                    <field name="remote_printer_ids" />
+                    <field name="remote_printer_ids" colspan="2" />
```
etobella: I think this does nothing. The other change has sense, but as we are inside a page without a group, we don't need to apply colspan here.

etobella (changes requested):

remote_report_to_printer/views/res_remote_views.xml
kluna1998: Changes done! 👍🏽 Thx

etobella (approved):
LGTM The error is related to translation

luisDIXMIT (approved):
LGTM!

PR #390, opened 2025-02-05
[18.0][MIG] printer_zpl2: Migration to 18.0

vvrossem:


printer_zpl2/models/printing_label_zpl2.py
```diff
@@ -64,7 +64,6 @@ class PrintingLabelZpl2(models.Model):
     action_window_id = fields.Many2one(
         comodel_name="ir.actions.act_window",
         string="Action",
-        readonly=True,
```
ivantodorovich: Why removing this? 🤔

printer_zpl2/wizard/wizard_import_zpl2.py
```diff
@@ -345,7 +345,9 @@ class WizardImportZPl2(models.TransientModel):
     _description = "Import ZPL2"
 
     label_id = fields.Many2one(
-        comodel_name="printing.label.zpl2", string="Label", required=True, readonly=True
+        comodel_name="printing.label.zpl2",
+        string="Label",
+        required=True,
```
ivantodorovich: I wouldn't remove readonly here

ivantodorovich (approved):

printer_zpl2/models/printing_label_zpl2.py
vvrossem: I've moved that attribute in the form view

printer_zpl2/wizard/wizard_import_zpl2.py
vvrossem: Agreed, it doesn't make sense to move it to the view definition within a wizard

printer_zpl2/models/printing_label_zpl2.py
vvrossem: > If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead.

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

ivantodorovich: I fail to see how this would make sense here (or in most other cases, tbh)
Here, the field is supposed to be set/unset by the actions `create_action` / `unlink_action`, for example


@pedrobaeza : I think you wrote this guideline, right? we'd appreciate your insight here 🙏🏻

pedrobaeza: I wrote it, because it was arisen in several PRs, and it's OK to remove in general `readonly=True` attributes from fields and move them to views. The text is already saying about that: if you need to import by whatever reason records from that model, you won't be able to import the fields marked as `readonly=True` in the fields, reducing the possibilities of the import because they can't be imported and leading to imported records with half of the data. You can't assume that they are going to be computed/filled somehow on create/write, because that's not the usual thing. That said, the sentence is correct because it's not forbidding it, so you judge if it makes sense to keep the `readonly=True` or not.

yankinmax (approved):

njeudy (approved):
Test in my production install, and rapid code review

dynapps-carlos (approved):
Reviewed and tested locally, worked fine as looks good to me!

PR #397, opened 2025-03-19
[16.0][REF] pingen_env: Leverage server-env mixins

grindtildeath:
Depends on:
 - [x] https://github.com/OCA/server-env/pull/232

### DIFF
--- a/pingen_env/models/res_company.py
+++ b/pingen_env/models/res_company.py
@@ -1,46 +1,22 @@
 # Copyright 2023 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
-from odoo import api, fields, models
-
-from odoo.addons.server_environment import serv_config
+from odoo import models
 
 
 class ResCompany(models.Model):
-    _inherit = ["res.company"]
+    _name = "res.company"
+    _inherit = ["res.company", "server.env.techname.mixin"]
 
-    pingen_clientid = fields.Char(
-        compute="_compute_pingen_env", required=False, readonly=True
-    )
-    pingen_client_secretid = fields.Char(
-        compute="_compute_pingen_env", required=False, readonly=True
-    )
-    pingen_organization = fields.Char(
-        compute="_compute_pingen_env", required=False, readonly=True
-    )
-    pingen_staging = fields.Boolean(
-        compute="_compute_pingen_env", required=False, readonly=True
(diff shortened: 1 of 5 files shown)


sebalix (approved):

cyrilmanuel (approved):

pingen_env/models/res_company.py
```diff
@@ -1,46 +1,22 @@
...
-        compute="_compute_pingen_env", required=False, readonly=True
-    )
-    pingen_staging = fields.Boolean(
-        compute="_compute_pingen_env", required=False, readonly=True
-    )
-    pingen_webhook_secret = fields.Char(
-        compute="_compute_pingen_env", required=False, readonly=True
-    )
+    _name = "res.company"
+    _inherit = ["res.company", "server.env.mixin", "server.env.techname.mixin"]
```
simahawk: FTR using the techname mixin implies that a techname is set on existing records and the env configuration must be updated. As a fallback, I would say that we need a migration script to automatically set the techname = to the current name. Tech name is normalized, so in the script it should be set via SQL if the original name is not normalized.

@grindtildeath am I wrong?

simahawk (changes requested):

pingen_env/models/res_company.py
grindtildeath: You're right on the techname, but there's no way a migration script is enough to have a "transparent" migration, as server environment values will have to be manually updated to reflect the techname anyway.

I'm not sure what's the best way to handle that though...

simahawk: I bet so far it relied on `name`. Just set the techname = name by default and everything should work fine.
Yet, as I mentioned, you should use an sql to update the name and bypass the normalization.
This way, you can avoid to touch the env conf. At least on the 1st upgrade.

grindtildeath: Actually this will still use the `name` field as `_server_env_section_name_field` :confused:

grindtildeath:
cf https://github.com/OCA/server-env/pull/232

grindtildeath:
I added a migration script to set the tech_name = name, in case it's not defined through another module yet. I also added a warning in the logs.

@simahawk Can we please merge the server-env PR and this one? :pray:

grindtildeath:
ping @simahawk

simahawk (approved):

PR #398, opened 2025-04-02
[15.0][IMP] base_report_to_printer: Add fast printer selector

CarlosRoca13:
Sometimes, a user may be working from different devices (if they're an operator) or moving around the company with a tablet and needs to print using the nearest printer, without losing focus on their current task.

With these changes, we allow them to quickly switch printers without having to access their user profile.

Here you have an example of how it works:
printer_selector

cc @Tecnativa TT55759

### DIFF
--- a/base_report_to_printer/README.rst
+++ b/base_report_to_printer/README.rst
@@ -89,6 +89,9 @@ directive to "none" (or some other list of values which does not include
 "job-name") , and reload the server. See `cupsd.conf(5)
 <https://www.cups.org/doc/man-cupsd.conf.html>` for details.
 
+If you want to see the button in the top bar to set the user's printer, you need
+to have the "Show printer button on navbar" group.
+
 Usage
 =====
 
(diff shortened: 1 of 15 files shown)


carlosdauden (approved):

sergio-teruel (approved):
Great!!!

pedrobaeza (commented):
Please hide the icon under a security group, as there will be a lot of users that don't care about their "default printer", and having that icon in such prominent space is too much IMO. We have to also think in a mobile display, where the width is very limited.

CarlosRoca13:
Added the `Show printer button on navbar` group to determine which users should see the button.

pedrobaeza:
Please add admin user to this new group for not having to make any setup to test it in runboat.

CarlosRoca13:
Done @pedrobaeza

pedrobaeza (approved):
/ocabot merge minor

PR #404, opened 2025-05-12
[15.0][FIX] base_report_to_printer: Available printers when CUPS can't be connected

carlosdauden:
The printers of a CUPS are still available even if no connection to the server can be established.

With this change we ensure that the status of the printers becomes unavailable.

@Tecnativa

### DIFF
--- a/base_report_to_printer/models/printing_server.py
+++ b/base_report_to_printer/models/printing_server.py
@@ -98,45 +98,50 @@ class PrintingServer(models.Model):
 
         res = True
         for server in servers.with_context(active_test=False):
-            connection = server._open_connection(raise_on_error=raise_on_error)
-            if not connection:
-                server.printer_ids.write({"status": "server-error"})
-                res = False
-                continue
-
-            # Update Printers
-            printers = connection.getPrinters()
-            existing_printers = {
-                printer.system_name: printer for printer in server.printer_ids
-            }
             updated_printers = []
-            for name, printer_info in printers.items():
-                printer = self.env["printing.printer"]
-                if name in existing_printers:
-                    printer = existing_printers[name]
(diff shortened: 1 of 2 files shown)


carlosdauden:
ping @sergio-teruel @CarlosRoca13

base_report_to_printer/models/printing_server.py
```diff
@@ -98,41 +98,43 @@ def update_printers(self, domain=None, raise_on_error=False):
...
+                    # empty names.
+                    if not printer_values.get("name") and not printer.name:
+                        printer_values["name"] = name
+                    if not printer:
+                        printer_values["system_name"] = name
+                        printer.create(printer_values)
+                    elif printer_values:
+                        printer.write(printer_values)
+            except Exception:
+                _logger.warning(_("Connection failed with %s", server.name))
```
rousseldenis: A little bit more context

```suggestion
                _logger.warning(_("Connection failed with printing server %s", server.name))
```

base_report_to_printer/models/printing_server.py
```diff
@@ -98,41 +98,43 @@ def update_printers(self, domain=None, raise_on_error=False):
...
             # Set printers not found as unavailable
             server.printer_ids.filtered(
```
rousseldenis: IMHO, with server error connection, printers status should become `server-error` and not `unavailable` as this is a server problem and not a printer connection problem

carlosdauden:
@rousseldenis changes done, thanks for your review

rousseldenis (approved):
LGTM even if function should be enhanced with some clever splits.

pedrobaeza:
Is this still draft? Please check CI.

PR #406, opened 2025-06-16
[16.0][FIX] base_report_to_printer: Print file only after commit

lmignon:
This is necessary to avoid printing documents when the process for which they are to be printed fails.

### DIFF
--- a/base_report_to_printer/models/printing_printer.py
+++ b/base_report_to_printer/models/printing_printer.py
@@ -164,7 +164,7 @@ class PrintingPrinter(models.Model):
             context = self.env.context
             uid = self.env.uid
             printer_id = self.id
-            report_id = report.id
+            report_id = report.id if report else None
 
             @self.env.cr.postcommit.add
             def _launch_print_file():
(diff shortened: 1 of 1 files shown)


base_report_to_printer/models/printing_printer.py
```diff
@@ -153,7 +154,40 @@ def print_document(self, report, content, **print_opts):
...
+            lambda: os.remove(file_name) if os.path.exists(file_name) else None
+        )
+
+        test_mode = getattr(threading.current_thread(), "testing", False)
+        if not test_mode:
+            dbname = self.env.cr.dbname
+            context = self.env.context
+            uid = self.env.uid
+            printer_id = self.id
+            report_id = report.id
```
jbaudoux:
```suggestion
            report_id = report.id if report else None
```

jbaudoux (changes requested):
It fails if report is not a recordset, for instance when the method is called in rpc

lmignon:
> It fails if report is not a recordset, for instance when the method is called in rpc

Thank you for the fix.

jbaudoux (approved):

rousseldenis (approved):

PR #409, opened 2025-07-05
[18.0][FIX] base_report_to_printer: removal of messageIsHtml

theangryangel:
This does not appear to exist and this results in a client side error sometimes.

Also cleared the eslint warning regarding ordering of imports.

I've also ensured that exceptions are logged, because that would've actually helped me with my real problem tonight.

### DIFF
--- a/base_report_to_printer/models/ir_actions_report.py
+++ b/base_report_to_printer/models/ir_actions_report.py
@@ -5,6 +5,7 @@
 # Copyright (C) 2013-2014 Camptocamp (<http://www.camptocamp.com>)
 # Copyright 2024 Tecnativa - Sergio Teruel
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+import logging
 import threading
 from time import time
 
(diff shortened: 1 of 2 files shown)


base_report_to_printer/models/ir_actions_report.py
```diff
@@ -118,6 +121,8 @@ def behaviour(self):
                     "unavailable",
                 ]
             except Exception:
+                if not self.env.context.get("skip_printer_exception"):
```
pedrobaeza, line 124: This is not related to the fix.

theangryangel: I'll happily raise a second PR... Unless you're saying that there's zero chance logging the exception won't be considered? 

For context, not logging an exception here was extremely frustrating in our case because a PDF would render, but direct print failed - but only because of the template - having the exception log made it stand out like a sore thumb.

theangryangel: Or are we specifically saying remove the context check?

pedrobaeza: I can't say about the change itself, but they are unrelated changes that have to be discussed separately, and you have to reason in the other PR why adding this.

pedrobaeza:
Superseded by #418

PR #410, opened 2025-07-09
[18.0][MIG] printing_auto_base: Migration to 18.0

florentx:
Migrate to Odoo 18.0


jbaudoux (changes requested):
where is the commits history?
did you followed https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0 ?

printing_auto_base/README.rst
```diff
@@ -0,0 +1,92 @@
+==================
```
jbaudoux, line 1: This file should have changed to .md file with pre-commit, isn't it?

florentx: I don't know ... It is `README.rst` in every other add-on.

IMHO it is not blocking on this PR, maybe something to fix/change in the tooling.

jbaudoux (approved):

printing_auto_base/README.rst
jbaudoux: yeah, it's all good :)

simahawk: it should happen automatically when you don't have `SKIP=oca-gen-addon-readme` env var set but it's only for the fragments. Here's it's ok!

simahawk (approved):

PR #417, opened 2025-08-29
[16.0][FIX] base_report_to_printer: do not download text report sent to printer

nicolas-delbovier-acsone:
In the case of a report of type "qweb-text" set to "Send to printer" (e.g. a zpl report), the pervious implementation was triggering a download of the report.

This is because inside the base odoo report handling function (odoo/addons/web/static/src/webclient/actions/action_service.js -> `_executeReportAction`), the code defaults to 'download' for text reports that have not been dealt with and current implementation only deals with "qweb-pdf" reports.
However, the supported types defined inside `REPORT_TYPES` are ["qweb-pdf", "qweb-text"].

Thus, the current approach now creates a new computed field `is_report_to_printer_action` (to replace the previous hard-coded solution) to deal with the logic of ensuring the report is of the expected type before triggering our custom action handler.

### DIFF
--- a/base_report_to_printer/models/ir_actions_report.py
+++ b/base_report_to_printer/models/ir_actions_report.py
@@ -35,11 +35,27 @@ class IrActionsReport(models.Model):
         string="Actions",
         help="This field allows configuring action and printer on a per " "user basis",
     )
+    is_report_to_printer_action = fields.Boolean(
+        compute="_compute_is_report_to_printer_action"
+    )
+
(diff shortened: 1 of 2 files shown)


lmignon (approved):

lmignon (approved):

rousseldenis (approved):
Seems legit. Thanks!

houssine78 (approved):
LGTM

PR #418, opened 2025-09-09
[18.0][FIX] report_print_send: messageIsHtml

david-s73:
When attempting to print a report from a batch, this error message appears, stating that the attribute “messageIsHtml” cannot be found.

image

There is already a PR open, but it is inactive and has more changes than it should. Here is the link to the PR.

https://github.com/OCA/report-print-send/pull/409


david-s73:
@pedrobaeza Upon reviewing the pr

pedrobaeza (approved):

miguel-S73 (approved):

PR #428, opened 2025-10-31
[18.0][FIX] base_report_to_printer: prevent error when trying to download the report manually after failing to connect to the printer.

carlos-lopez-tecnativa:
When the connection to the printer fails, an option is provided to the user to download the PDF. However, when the user clicks the button, an error is obtained. This commit fixes the error by removing the line, because in `doAction` the context is passed through `additionalContext`.

@Tecnativa @pedrobaeza @sergio-teruel  @CarlosRoca13  could you please review this?

image

### DIFF
--- a/base_report_to_printer/static/src/js/qweb_action_manager.esm.js
+++ b/base_report_to_printer/static/src/js/qweb_action_manager.esm.js
@@ -70,7 +70,6 @@ async function cupsReportActionHandler(action, options, env) {
                                     force_print_to_client: true,
                                     must_skip_send_to_printer: true,
                                 };
-                                env.services.user.updateContext(context);
                                 await env.services.action.doAction(
                                     {type: "ir.actions.report", ...action},
                                     {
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

sergio-teruel (approved):

pedrobaeza:
@carlos-lopez-tecnativa again the CI problem.

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa again the CI problem.

Yes, the error is the same. The log displays a `429` error, and according to the API documentation https://labelary.com/service.html, the limits are:

`Requests per second: 3`

So, when running OCB and Odoo as parallel jobs, I imagine they reach this limit of 3 requests per second.
It’s necessary to add a sleep or patch the methods. What do you think?

image

image

pedrobaeza:
Well, in this case, I wouldn't test the external API due to the existing constraints.

carlos-lopez-tecnativa:
> Well, in this case, I wouldn't test the external API due to the existing constraints.

Done, please try again.

PR #429, opened 2025-10-31
[18.0]FIX] base_report_to_printer: return a close action when the option is specified

carlos-lopez-tecnativa:
Many modules can return the key `close_on_report_download` in the context. In the `web` module, this option is expected and handled accordingly. However, this module ignored the option, causing the wizard not to close when a report is generated from an action with target 'new' (typically a wizard). After this commit, the wizard closes correctly.

https://github.com/odoo/odoo/blob/401540aaa8e42243f4504c594d477330295ddee0/addons/web/static/src/webclient/actions/action_service.js#L1326-L1330

Example in the `delivery_package_number` module: https://github.com/OCA/delivery-carrier/blob/12e690c7b912073c2aa54be07868d84cbe24f438/delivery_package_number/wizard/stock_number_package_validate_wiz.py#L65

@Tecnativa @pedrobaeza @sergio-teruel  @CarlosRoca13  could you please review this?

### DIFF
--- a/base_report_to_printer/static/src/js/qweb_action_manager.esm.js
+++ b/base_report_to_printer/static/src/js/qweb_action_manager.esm.js
@@ -24,6 +24,15 @@ async function cupsReportActionHandler(action, options, env) {
                 env.services.notification.add(_t("Successfully sent to printer!"), {
                     type: "success",
                 });
+                const {onClose} = options;
+                if (action.close_on_report_download) {
+                    return env.services.action.doAction(
+                        {type: "ir.actions.act_window_close"},
+                        {onClose}
+                    );
+                } else if (onClose) {
+                    onClose();
+                }
                 return true;
                 // In case of exception during the job, we won't get any response. So we
                 // should flag the exception and notify the user


pedrobaeza (approved):

carlos-lopez-tecnativa:
The error in the CI is unrelated to the change. I see that the `printer_zpl2` module tries to make a request to an external URL, so the response is not as expected, but I think it’s a fortuitous case.

https://github.com/OCA/report-print-send/blob/91e7cc06fcae823956d1dc4d07ab5a3df5c4c095/printer_zpl2/models/printing_label_zpl2.py#L528

sergio-teruel (approved):

pedrobaeza:
@carlos-lopez-tecnativa can you please check the error in the CI that is consistently reproduced only on one of the flavors?

carlos-lopez-tecnativa:
> @carlos-lopez-tecnativa can you please check the error in the CI that is consistently reproduced only on one of the flavors?

I tried to reproduce the error locally, but the test passed correctly. Could you please try again?
image
image

pedrobaeza:
Trying again

/ocabot merge patch

PR #430, opened 2025-11-01
printer_zpl2: fix label._get_record method

kirca:
use model.browse() to check if the record exists.
model.search() by `id` doesn't work.


printer_zpl2/models/printing_label_zpl2.py
```diff
@@ -454,8 +454,8 @@ def import_zpl2(self):
     def _get_record(self):
         self.ensure_one()
         Obj = self.env[self.model_id.model]
-        record = Obj.search([("id", "=", self.record_id)], limit=1)
-        if not record:
+        record = Obj.browse(self.record_id)
```
pedrobaeza:
```suggestion
        record = Obj.browse(self.record_id).exists()
```

PR #365, opened 2024-09-17
[17.0][IMP] base_report_to_printer : add output_tray

ecino:
This pull request refactors the printer tray handling in the `base_report_to_printer` module to support separate input and output trays, improving flexibility.

### Model and Field Refactoring

* Split the generic `printing.tray` model into two distinct models: `printing.tray.input` and `printing.tray.output`, and updated all related fields in `ir.actions.report`, `printing.printer`, `printing.report.xml.action`, and `res.users` to use these new models for input and output trays. 
* Updated onchange and default behaviour methods to reset and return values for both input and output trays, ensuring correct handling when printer or tray selections change.

### Printer Synchronization and Options

* Modified the printer update logic to separately synchronize input and output trays from CUPS, and added new methods for setting printer options for both tray types.

### Security and Access Control

* Updated security rules to grant appropriate access to the new input and output tray models, including separate permissions for users and managers.

### Tests and Backwards Compatibility

* Refactored tests to use the new input/output tray fields and models, ensuring all behaviour checks are updated and remain valid.

### Miscellaneous

* Updated the module version to `17.0.1.1.2` to reflect these changes.

PR #431, opened 2025-11-19
[18.0][IMP] base_report_to_printer : add output_tray

ecino:
This is the same as #365 adapted for v18.

V17 has been accepted but weird merging errors have prevented the merge to happen. I hope this one will be more successful.


ecino:
@pedrobaeza @sergio-teruel Those are the same changes as approved in v17 adapted for v18. I also added a commit for adapting the submodules to those changes.

pedrobaeza (commented):
Now you have to take into account in the migration scripts if someone is coming from 17 with the change already done, so they won't have that the old model in its DB.

ecino:
@pedrobaeza Good point! I updated the migration script.

pedrobaeza (approved):

ecino:
Does this (and v19 changes) get higher priority now that the version 17 is merged? To be honest I didn't test what would happen when migrating from version 17 to 18 when the version 17 has the new fields for output trays and version 18 don't have them.

pedrobaeza:
Thanks for your honesty. Yes, migration scripts are difficult to test. Doing a code review, they seem correct, but only after performing such migration, we can be sure.

Let's assume the risk and sync all the branches:

/ocabot merge minor

PR #433, opened 2025-12-17
[18.0][FIX] base_report_to_printer: neutralization scripts

ferran-S73:

### DIFF
--- a/base_report_to_printer/views/ir_actions_report.xml
+++ b/base_report_to_printer/views/ir_actions_report.xml
@@ -20,9 +20,27 @@
                         />
                     </group>
                     <separator string="Specific actions per user" />
-                    <field name="printing_action_ids" />
+                    <field name="printing_action_ids">
+                        <list editable="bottom" decoration-muted="active == False">
+                            <field name="user_id" />
+                            <field name="action" />
+                            <field name="printer_id" />
+                            <field
+                                name="printer_input_tray_id"
+                                invisible="not printer_id"
+                            />
+                            <field
+                                name="printer_output_tray_id"
+                                invisible="not printer_id"
+                            />
+                            <field name="active" optional="hide" />
+                        </list>
+                    </field>
                 </page>
             </page>
         </field>
+    </record>
+    <record model="ir.actions.act_window" id="base.ir_action_report">
+        <field name="context">{'active_test': False}</field>
     </record>
 </odoo>
--- a/base_report_to_printer/data/neutralize.sql
+++ b/base_report_to_printer/data/neutralize.sql
@@ -1,2 +1,4 @@
 update printing_server set active=false;
 update printing_printer set active=false;
+update printing_report_xml_action set active=false;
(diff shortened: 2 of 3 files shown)


pedrobaeza (commented):
Please remove the version from the commit message, and about the functionality, it's interesting, but please add all the stuff for the active field like banner.

ferran-S73:
@pedrobaeza removed version from commit message + rebased to resolve conflicts. As per the banner, are you talking about the ribbon with the text "archived" that is shown in some views? If that is the case I can't add it here since there is no actual form view for the `printing.report.xml.action` model.

pedrobaeza:
OK, if all is handled from the list view, then it should be handled in a similar way to crons, that all are shown through a default filter, and an extra one should be to have a pill to identify the active/inactive ones in a column.

Reyes4711-S73 (approved):
LGTM

miquelalzanillas (approved):
LGMT

Reyes4711-S73:
@OCA/tools-maintainers Please, can you review/merge this PR?

pedrobaeza:
@ferran-S73 maybe you can do the `active_test=False` context key in the action and color differentiation for completing the change?

ferran-S73:
Hi @pedrobaeza , currently there's no window action nor a list view with all the printing_report_xml_action, it's all controlled via the ir_actions_report form view

pedrobaeza:
Then the context and decoration should be applied to the o2m field. The aim is to be able to see that the records are archived. If not, you won't see anything and doubt if this has been neutralized, or directly they don't exist.

miquelalzanillas:
Hello @ferran-S73 ,

I think @pedrobaeza is right about the visibility of an inactive action. 

image

Decoorating these rows it would give more security to the user that the actions are neutralized.

ferran-S73:
@pedrobaeza @miquelalzanillas aaah I see what you mean now, thanks for clarifying! I'll sure be adding those changes!

ferran-S73:
Hi again @pedrobaeza @miquelalzanillas changes added, I had to add the active_test context to the ir_actions_report window action in order for it to really work. I don't think it'll be a problem since the ir_actions_report does not have an active field.

base_report_to_printer/views/ir_actions_report.xml
```diff
@@ -20,9 +20,27 @@
                         />
                     </group>
                     <separator string="Specific actions per user" />
-                    <field name="printing_action_ids" />
+                    <field name="printing_action_ids">
+                        <list editable="bottom" decoration-muted="active == False">
```
pedrobaeza, line 24:
```suggestion
                        <list editable="bottom" decoration-muted="not active">
```

pedrobaeza (approved):
But let's not block for that nitpicking

/ocabot merge minor

pedrobaeza:
Please fw-port it to 19.

ferran-S73:
> Please fw-port it to 19.

done! https://github.com/OCA/report-print-send/pull/460

PR #434, opened 2025-12-18
[17.0][FIX] base_report_to_printer: neutralization scripts

Vicent-C6:

The second commits is for: 
    - The tests are designed to check the API response, but the problem is that two requests are made within a short time      interval. A mock would be appropriate if what was being checked was not the response from api.labelary. 
   
image


Reyes4711-S73 (approved):
LGTM

PR #443, opened 2026-01-22
[18.0][FIX] base_report_to_printer: Print double label

david-s73:
- These changes have been removed because we found several cases in which, instead of printing a single label, two labels were being printed. Removing the change means that only one label is printed again.

### DIFF
--- a/base_report_to_printer/static/src/js/qweb_action_manager.esm.js
+++ b/base_report_to_printer/static/src/js/qweb_action_manager.esm.js
@@ -3,7 +3,7 @@ import {markup} from "@odoo/owl";
 import {registry} from "@web/core/registry";
 
 async function cupsReportActionHandler(action, options, env) {
-    if (action.report_type === "qweb-pdf" || action.report_type === "qweb-text") {
+    if (action.report_type === "qweb-pdf") {
         const orm = env.services.orm;
 
         const print_action = await orm.call(


ferran-S73 (approved):

Reyes4711-S73 (approved):

Reyes4711-S73:
@OCA/tools-maintainers Please, can you review/merge this PR?

thomaspaulb:
I don't understand the fix though - I think the qweb-pdf and the qweb-text should both be supported because one can have for example PDF reports but also ZPL reports that are to be sent to the printer.

How does removing support for qweb-text solve a duplication issue?

Raul-S73:
@thomaspaulb Following an analysis of several production cases, we have identified that the previous modification unexpectedly interferes with the qweb-text report rendering flow, causing the print action to trigger twice.

Given the urgency of the reported tickets, our priority is to revert this anomalous behavior. We propose merging this fix to stabilize the service while we re-evaluate the auto-printing strategy in a controlled testing environment.

Previous modification:
https://github.com/OCA/report-print-send/commit/968036706a33188137416ead130e6f5e0b04061d

simahawk:
I agree w/ @thomaspaulb . Have you tried to replicate on Runboat?

Technical notes: you should use `git revert $sha`. However, I see that the same commit changed the version of the module: that should not be done. Regarding the commit of this PR: pls do not add the Odoo version. It's not relevant and makes fwd/backport akward.

Raul-S73:
Hi @simahawk , yes, I’ve tested it and confirmed what we already knew...

I’m attaching the images so you can see how it looks currently in production versus with this PR.
Reverting wasn't an option because, if you check the PR I mentioned in my previous comment, there are other changes included beyond this one.

Regards,

Currently in production:
image

With this changes:
image

thomaspaulb:
Ok you did the functional test but if you would for example place a breakpoint [here](https://github.com/odoo/odoo/blob/c1493cbf1b66295a2663ed633def560c864f7251/addons/web/static/src/webclient/actions/action_service.js#L1297), what is actually happening? Is the handler somehow returning `False`, so it will also try other handlers? Or is the ZPL sent to the printer -and- still being downloaded and somehow sent to the printer again via the local computer? Or does the handler call the backend print action twice?

thomaspaulb:
> Reverting wasn't an option because, if you check the PR I mentioned in my previous comment, there are other changes included beyond this one.

@Raul-S73 Which other changes were you talking about?

I can revert the SHA just fine, only had to manually resolve the version conflict:

image

thomaspaulb:
More proper PR: https://github.com/OCA/report-print-send/pull/445

Does the `time.sleep(3)` addition to the unit test also need reverting?

Raul-S73:
Hello @thomaspaulb ,

We do want to keep those changes you mentioned. Apologies, as I provided the commit link instead of the PR link. The change you uploaded is exactly the same as ours; couldn't you merge this one?

thomaspaulb:
Well, I don't know how important it is to have a proper "git revert"-style commit message like @simahawk suggested, or not. It gives that slightly better commit history. So merging that one, instead of this.

PR #445, opened 2026-01-31
[18.0] Revert "base_report_to_printer: Print ZXPL without downloading"

thomaspaulb:
This reverts commit 968036706a33188137416ead130e6f5e0b04061d from https://github.com/OCA/report-print-send/pull/442


thomaspaulb:
/ocabot merge patch

fasttracking because content same as https://github.com/OCA/report-print-send/pull/443 which was approved, this just has the proper "Revert" commit message

PR #449, opened 2026-03-12
[19.0][IMP] base_report_to_printer : add output_tray

ecino:
This is the same as https://github.com/OCA/report-print-send/pull/365 adapted for v19.

V17 has been accepted but weird merging errors have prevented the merge to happen. I hope this one will be more successful.


ecino:
@pedrobaeza @sergio-teruel And now I also migrated those changes for v19. Would you be willing to review ? This way we will have those additions in the latest version as well which I believe brings more flexibility for printing jobs, which is especially helpful when using big office printers.

ecino:
I also adapted the migration script here.

pedrobaeza (approved):
Let's go with this one as well.

/ocabot merge minor

PR #453, opened 2026-04-16
[17.0][ADD] base_report_to_label_printer : copied from 18.0

luffah:
This module is the direct copy from branch 18.0.
I just changed 18.0 by 17.0, installed and tested.


pedrobaeza:
You have to preserve commit history.

luffah:
(i'm not sure how to correct that)
Given this is an odd version of Odoo and that the module is not a key functionnality. I remove this PR.

PR #460, opened 2026-05-18
[19.0][FIX] base_report_to_printer: neutralization scripts

eugenios73:
**fixes impemented in this v18.0 PR https://github.com/OCA/report-print-send/pull/433**


ferran-S73 (approved):

Reyes4711-S73 (approved):
LGTM

pedrobaeza (commented):
And the neutralization script?

ferran-S73:
> And the neutralization script?

Already merged in a previous commit https://github.com/OCA/report-print-send/blame/19.0/base_report_to_printer/data/neutralize.sql

pedrobaeza (approved):
/ocabot merge patch

PR #463, opened 2026-06-05
[18.0][IMP] base_report_to_printer: Add fast printer selector

CarlosRoca13:
Sometimes, a user may be working from different devices (if they're an operator) or moving around the company with a tablet and needs to print using the nearest printer, without losing focus on their current task.

With these changes, we allow them to quickly switch printers without having to access their user profile.

FW-PORT of: https://github.com/OCA/report-print-send/pull/398

cc @Tecnativa TT62927


sergio-teruel (approved):

eduezerouali-tecnativa (approved):
Tested on runboat

PR #465, opened 2026-06-23
[19.0][FIX] base_report_to_printer: make action_type selection callable

niboo-vli:
The action_type Selection referenced the _available_action_types property object directly in the class body, so field.selection held a property instance. fields_get then raised 'TypeError: property object is not iterable' whenever the model was introspected (opening the view, Studio).

Wrap the property access in a lambda, like the two other selection fields of the module already do.

### DIFF
--- a/base_report_to_printer/models/printing_action.py
+++ b/base_report_to_printer/models/printing_action.py
@@ -22,5 +22,7 @@ class PrintingAction(models.Model):
 
     name = fields.Char(required=True)
     action_type = fields.Selection(
-        selection=_available_action_types, string="Type", required=True
+        selection=lambda self: self._available_action_types,
+        string="Type",
+        required=True,
     )


ivantodorovich (approved):
LG

Ricardoalso (approved):

base_report_to_printer/models/printing_action.py
```diff
@@ -22,5 +22,7 @@ def _available_action_types(self):
 
     name = fields.Char(required=True)
     action_type = fields.Selection(
-        selection=_available_action_types, string="Type", required=True
+        selection=lambda self: self._available_action_types,
```
pedrobaeza, line 25: Shouldn't it be this?

```suggestion
        selection=lambda self: self._available_action_types(),
```

pedrobaeza: Oh, wait, it was converted to a property...

pedrobaeza (approved):
/ocabot merge patch

PR #466, opened 2026-06-29
[19.0][MIG] printer_zpl2

micheledic:


micheledic:
@pedrobaeza can you check?

micheledic:
@pedrobaeza can you check ?

pedrobaeza:
Sorry, others should review. I can just handle general management, but I'm not using this module.

micheledic:
@MiquelRForgeFlow can you check ?

printer_zpl2/migrations/19.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,24 @@
+from openupgradelib import openupgrade
+
+# 18.0 shipped the label list view and its act_window.view link under "tree"
+# xml ids. 19.0 renames both to "list" and adds the unique constraint
+# act_window_view_unique_mode_per_action on (act_window_id, view_mode). The
+# existing act_window.view row already holds "list" (converted on upgrade), so
+# reloading the data under the new xml id would insert a duplicate list row and
+# the constraint would abort the upgrade. Rename the xml ids up front so the
+# reload updates the existing rows in place.
+_XMLID_RENAMES = [
```
MiquelRForgeFlow: remove the migration script. First, if needed, it should be done in 18.0 branch instead, as it is where is was done the rename. But it's not needed, as this is noupdate=0 data.

micheledic: > remove the migration script. First, if needed, it should be done in 18.0 branch instead, as it is where is was done the rename. But it's not needed, as this is noupdate=0 data.

I migrated directly from 16.0 to 19.0, and when updating printer_zpl2 I got an error because the old XML ID already existed for that view type, so the constraint was raised.

That’s why I added the migration scripts. However, if you think they should be on 18.0 PR, I’ll remove them from the PR.

MiquelRForgeFlow: Yes, remove them from this PR. You can do another PR for 18.0 to add scripts. But it's strange, anyway, because this kind of data is noupdate=0, so you shouldn't have a problem.

micheledic: > Yes, remove them from this PR. You can do another PR for 18.0 to add scripts. But it's strange, anyway, because this kind of data is noupdate=0, so you shouldn't have a problem.

The reason is that new record with new xmlid is added before the remove of the old xmlid that are not present on the module. The old xmlid is converted by the odoo upgrade scripts to view_type "list" . however i removed it

MiquelRForgeFlow: Oh, right, I understand. Please add a PR to v18.

MiquelRForgeFlow (approved):

micheledic:
@MiquelRForgeFlow merge?

giovannigd (approved):

ivantodorovich (approved):

ivantodorovich:
This should be ready to merge

micheledic:
> This should be ready to merge

@MiquelRForgeFlow

MiquelRForgeFlow:
@micheledic I don't have merge powers here.

micheledic:
@pedrobaeza sorry for ping, can you merge? it is approved.
Where can i see who has merge powers for each repository ?

PR #470, opened 2026-07-13
[17.0][IMP] Make 'server_id' mandatory only for CUPS printers

miquelalzanillas:
With this changes, 'server_id' will only be mandatory for printers with 'cups' backend. Right now any printer can be created without set 'cups' as backend because of this limitation.

@pedrobaeza @cubells @bernatobrador

### DIFF
--- a/base_report_to_printer/models/printing_printer.py
+++ b/base_report_to_printer/models/printing_printer.py
@@ -12,7 +12,8 @@ import logging
 import os
 from tempfile import mkstemp
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Is this adding a new field?

miquelalzanillas:
> Is this adding a new field?

No, the commit was wrong. Now is fixed, sorry.

cubells (approved):

pedrobaeza (commented):
I think you should put the field as required in the view if the backend is that one.

miquelalzanillas:
Done also here.

Colud you please check again @rousseldenis ?

pedrobaeza:
In 17, you need to add as invisible the `backend` field.

pedrobaeza (approved):

BernatObrador (approved):
LGTM

rousseldenis (approved):

PR #471, opened 2026-07-13
[18.0][FWD][IMP] Make 'server_id' mandatory only for CUPS printers

miquelalzanillas:
Forward port of https://github.com/OCA/report-print-send/pull/470

@cubells @bernatobrador

### DIFF
--- a/base_report_to_printer/models/printing_printer.py
+++ b/base_report_to_printer/models/printing_printer.py
@@ -12,7 +12,8 @@ import logging
 import os
 from tempfile import mkstemp
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
It seems the commit message doesn't contain the module name.

rousseldenis:
I don't like this implementation. Prefer #470 with a constraint on backend side instead.

If we have several different backends, the view coherence is not guaranteed.

miquelalzanillas:
I'm agree with @rousseldenis. So I think I will back to it. Are you agree @pedrobaeza ?

pedrobaeza:
Yes, of course, I think my words in https://github.com/OCA/report-print-send/pull/470#pullrequestreview-4685457570 were misunderstood. I didn't say to swap the method, but to add it to the view as well (sorry if I wasn't specific enough).

miquelalzanillas:
No problem. Done here.

pedrobaeza (approved):

BernatObrador (approved):
LGTM

rousseldenis (approved):

PR #475, opened 2026-08-25
[19.0] [FIX] base_report_to_printer: restore the default printer hook 

ivantodorovich:
This override didn't work since b75c6b0, where the hook that was originally introduced in a27fced got removed: 
- https://github.com/OCA/report-print-send/blob/77587409dcb852b702bdfb80e9ecc7d0b3d903f8/base_report_to_label_printer/models/ir_actions_report.py#L12-L15

It was not caught by tests

### DIFF
--- a/base_report_to_printer/models/ir_actions_report.py
+++ b/base_report_to_printer/models/ir_actions_report.py
@@ -76,12 +76,23 @@ class IrActionsReport(models.Model):
             serializable_result["action"] = "client"
         return serializable_result
 
+    def _get_user_default_printer(self, user):
+        """Return the printer ``user`` prints this report on by default.
+
+        Hook for modules that route some reports to another printer, such as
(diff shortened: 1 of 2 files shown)


Ricardoalso (approved):

base_report_to_printer/models/ir_actions_report.py
```diff
@@ -76,12 +76,16 @@ def print_action_for_report_name(self, report_name):
             serializable_result["action"] = "client"
         return serializable_result
 
+    def _get_user_default_printer(self, user):
+        return user.printing_printer_id
```
pedrobaeza, line 87: Add a docstring for the method stating that it can be overriden for changing the printer in which cases..

ivantodorovich: done!

pedrobaeza (approved):
/ocabot merge patch

PR #478, opened 2026-09-08
[19.0][FIX] base_report_to_printer: forward doc_format, and guard neutralize.sql

micheledic:
## print_document drops doc_format

`print_document` takes `doc_format` as a named argument and forwards only `**kwargs`, so the value never reaches `print_file`, `print_options` or `_set_option_doc_format`. A caller asking for a raw job gets a job with no options at all.

`printer_zpl2` prints that way:

    printer.print_document(report=None, content=label_contents, doc_format="raw")

Without the raw option CUPS hands the file to the queue driver. On a Zebra queue carrying a ZPL PPD the driver renders the ZPL as text, so the label comes out as a page of `^XA^FO...` commands instead of the label itself.

Measured on a 19.0 database with `print_file` instrumented:

| | kwargs reaching print_file | CUPS options |
|---|---|---|
| 16.0 | `{'doc_format': 'raw'}` | `{'raw': 'True'}` |
| 19.0 before this branch | `{}` | `{}` |

The 16.0 signature was `print_document(self, report, content, **print_opts)`, so the format travelled inside `print_opts`. The refactor that extracted the CUPS backend, b75c6b0, turned it into a parameter and left the forwarding
behind. 
A test pins it: `print_document(..., doc_format="raw")` reaches `print_file`
and resolves to `{'raw': 'True'}`.

## neutralize.sql aborts when the active column is missing

Neutralization runs on the restored dump, before the modules update, so a database upgraded from an earlier version has no `active` column on `printing_report_xml_action` yet: the module adds it on its own update. The
statement then raises

    ERROR: column "active" of relation "printing_report_xml_action" does not exist

and the build stops with the printers left enabled, the opposite of what neutralization is for.

The update now runs inside a `DO` block that reads `information_schema.columns` first. Checked both ways on a 19.0 database: with the column the three statements apply, without it the guarded one is skipped and nothing raises.

issue: https://github.com/OCA/report-print-send/issues/479

### DIFF
--- a/base_report_to_printer/README.rst
+++ b/base_report_to_printer/README.rst
@@ -164,34 +164,34 @@ Usage
 
 Guidelines for use:
 
-- To use a specific printing backend (e.g. CUPS), make sure the
-  corresponding module (such as ``base_report_to_printer_cups``) is
-  installed and configured.
-- To print a report on a specific printer or tray, you can configure
-  defaults in *Settings > Printing > Reports*.
-- To define user-specific behaviour, go to *Settings > Printing >
-  Reports* and configure *Specific actions per user*.
-- Each user can also select a default action, printer or tray in their
-  *Preferences*.
-- When no tray is configured for a report or a user, the default tray
-  defined by the printing backend (e.g. the CUPS server) will be used.
+-  To use a specific printing backend (e.g. CUPS), make sure the
+   corresponding module (such as ``base_report_to_printer_cups``) is
+   installed and configured.
+-  To print a report on a specific printer or tray, you can configure
+   defaults in *Settings > Printing > Reports*.
+-  To define user-specific behaviour, go to *Settings > Printing >
+   Reports* and configure *Specific actions per user*.
+-  Each user can also select a default action, printer or tray in their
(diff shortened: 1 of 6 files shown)


ivantodorovich:
Why are you neutralizing before the upgrade?

Every module with a neutralization script has this same problem, so IMO this fix doesn't belong to the script itself (else, we'd need the same everywhere!)

To workaround it, we always upgrade before neutralize

micheledic:
> Why are you neutralizing before the upgrade?
> 
> Every module with a neutralization script has this same problem, so IMO this fix doesn't belong to the script itself (else, we'd need the same everywhere!)
> 
> To workaround it, we always upgrade before neutralize

On odoo.sh the neutralize happens before the upgrade so goes in error

ivantodorovich:
IMHO it sounds like an odoo.sh problem.
I'm not fully against this pattern, but if we go with it then I'm afraid lots of modules would need it -- cause this can happen everywhere.
As a workaround, can't you create the missing column in Odoo.sh db before neutralize happens?

PR #482, opened 2026-09-23
[19.0] [IMP] printer_zpl2: labelary and test printing improvements/fixes

ivantodorovich:
Several improvements/fixes, each in its own commit


printer_zpl2/models/printing_label_zpl2.py
```diff
@@ -85,9 +87,36 @@ class PrintingLabelZpl2(models.Model):
         required=True,
         default="8dpmm",
     )
-    labelary_width = fields.Float(string="Width in mm", default=140)
+    labelary_width = fields.Float(
```
Ricardoalso: This requires a call to `_compute_labelary_width` in a post migration script

printer_zpl2/wizard/print_test.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+    Copyright 2026 Camptocamp SA (https://www.camptocamp.com).
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+<odoo>
+    <record id="view_wizard_zpl2_print_test_form" model="ir.ui.view">
+        <field name="name">wizard.zpl2.print_test.form</field>
```
Ricardoalso:
```suggestion
```

Ricardoalso (commented):
~~Missing compute trigger on historic data for the new computed `labelary_width` field~~

printer_zpl2/models/printing_label_zpl2.py
ivantodorovich: Hmm. Not sure. I rather keep the user already provided value and only trigger a recompute if they actually change the width

Ricardoalso: ~~Even if one of the 2 dependent fields (or both) are empty?~~

`@api.depends("width", "labelary_dpmm")`

**Edit**
Well there was a default 140, makes sense then

Ricardoalso (approved):

